PR #1, opened 2020-02-18
Makefile: Link to math lib

ilou89:
Hi,
I think the "-lm" should be added in order to use the math lib functions (sqrtf etc).

ps. I only ran the "make build_cli", could not ran "make build" yet because I'm currently missing the cunit framework

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ valgrind:
 
 make build_cli:
 	rm -f connxr
-	gcc -std=c99 -Wall -D TRACE_LEVEL=0 -o connxr src/operators/*.c src/*.c src/pb/onnx.pb-c.c src/pb/protobuf-c.c
+	gcc -std=c99 -Wall -D TRACE_LEVEL=0 -o connxr src/operators/*.c src/*.c src/pb/onnx.pb-c.c src/pb/protobuf-c.c -lm
 
 #memory leak stuff TODO:
 


alrevuelta:
Fair comment, are you using Linux? In macOS it works fine without `-lm`. I just enabled CI for pull requests, so if you create one more commit it should trigger all the tests (including `make build`). I will try to add other platforms such as Linux to the CI action, and to be honest, the `Makefile` needs some work also.
So can you commit again and see if the action is triggered?
Thanks!

ilou89:
Hi thanks for the info! Indeed, I tried it on Linux (Ubuntu). The CI seems fine (it passes in my forked repo), so feel free to accept this minor change :)

alrevuelta:
Just added Ubuntu to the CI. Thanks again!

alrevuelta:
Heads up! There seems to be an issue in Ubuntu with `maxpool.c` operator, since tests are failing. That makes also MNIST model to output wrong results.

ilou89:
I also tried to run the tiny_yolov2 model on ubuntu and I get segmentation fault. I didn't have time to look it further, when I have any news I'll let you know!

alrevuelta:
@ilou89 I just fixed the `maxpool.c` operator that was affecting `MNIST` model and probably `tiny_yolov2`. There was an issue with `strcmp` in Ubuntu when comparing if the padding was `SAME_UPPER`. I replaced it with `strncmp` so now we only check that exactly the first 10 characters are equal:

```c
if (!strncmp((char*)auto_pad->s.data, "SAME_UPPER", 10)){
  //...
}
```

So now `MNIST` model is working as expected but this haven't fixed `tiny_yolov2`. I have created an issue here to track this https://github.com/alrevuelta/cONNXr/issues/2 so lets move the conversation there.

Thanks!

PR #6, opened 2020-03-24
fixed type and unused warnings

nopeslide:


nopeslide:
not happy with this. needs a little more time.

PR #7, opened 2020-03-24
rewrote Makefile, restructured build dirs and simplified .gitignore

nopeslide:


Makefile
```diff
@@ -1,52 +1,162 @@
-all: clean build
-	./runtest $(ts) $(tc)
```
alrevuelta: Is it possible to preserve this functionality? For example if you run `make test_operators`, would be nice to accept an optional parameter to select the operator that you want to run i.e. `make test_operators tc=test_operator_maxpool_1d_default`. Same for `make test_models`

nopeslide: How about I introduce a `OPERATORS` variable next to the already existing `MODELS` one to constrain these tests?

nopeslide: > I have noted also something odd in the CI when running "run_benchmark". Macos is working as expected but ubuntu is only running 1 iteration for MNIST, hence not averaging any value.

This may be a dash/bash problem. bash expands `{1..10}` to something similar like `seq 10` produces. will replace it. good catch!

nopeslide: > How about I introduce a `OPERATORS` variable next to the already existing `MODELS` one to constrain these tests?

this will be one long list. Would introduce a special case if the `OPERATORS` and `MODELS` lists are empty to run the whole set.

alrevuelta:
Looks good to me!
I would remove the `test_` from `build/runtest onnxBackendSuite test_$OPERATOR; \` in the makefile, so that the operator to test has the exact name as in the onnx test backend and in `tests.c` (i.e. test_and_bcast4v4d) so it can be copy-pasted. But its a matter of taste.

Thanks!

PR #9, opened 2020-04-01
Make compilation free of warnings

nopeslide:


alrevuelta:
Nice! I saw you left the `Werror` flag commented in the [makefile](https://github.com/alrevuelta/cONNXr/blob/master/Makefile)
`# CFLAGS+=-Werror`

Can't we just uncomment that so we don't have to modify every `yml` file?

nopeslide:
> Nice! I saw you left the `Werror` flag commented in the [makefile](https://github.com/alrevuelta/cONNXr/blob/master/Makefile)
> `# CFLAGS+=-Werror`
> 
> Can't we just uncomment that so we don't have to modify every `yml` file?

We could, but it think its annoying if you just want to change stuff quick.
The CI should do it, but depending on what you are doing it may get into your way,
then you disable it, it shows as modified, you have to remind yourself before each commit and so on.
just a hassle in my opinion.

alrevuelta:
> We could, but it think its annoying if you just want to change stuff quickly.
> The CI should do it, but depending on what you are doing it may get into your way,
> then you disable it, it shows as modified, you have to remind yourself before each commit and so on.
> just a hassle in my opinion.

I expected it to be an always on flag, but if it might be modified, then you are right.

nopeslide:
> > We could, but it think its annoying if you just want to change stuff quickly.
> > The CI should do it, but depending on what you are doing it may get into your way,
> > then you disable it, it shows as modified, you have to remind yourself before each commit and so on.
> > just a hassle in my opinion.
> 
> I expected it to be an always on flag, but if it might be modified, then you are right.

And since you are still getting the warnings, we're not hiding stuff.
We may add multiple compile stages in the future and then `-Werror` gets annoying since subcomponents won't compile, because previous ones errored

PR #10, opened 2020-04-01
Operator generation from onnx

nopeslide:
- onnx as submodule
  - [x] freezed on latest version
  - [x] wrote Makefile handling onnx build
- generator script
  - [x] operator header
  - [x] operator doxygen documentation
  - [x] resolve function for operator type
  - [x] global operator stub
  - [x] resolve function for operator name
  - [x] stub for each type combination
  - [x] fallback to installed onnx instead of self-compiled
  - [x] operator subset
  - [x] domain subset
  - [x] version subset
  - [x] old operators
  - [x] ~type subsets~
    - removed, makes no sense to limit this in the generation step
  - [x] return enum
  - [x] domain specific header directory
  - [x] domain in operator name
  - [x] version in operator name
  - [x] domain specific implementation directory
  - [x] domain specific header filename
  - [x] domain specific implementation filename
  - [x] ~switch to replace weakref with hardcoded stub~
    - replaced alias with weak function symbol
  - [x] ~generate headerfiles which include whole domains~
    - mitigated with sets, no need to include whole domains

## Questions
- how to manage generated files?
  - should be generated with main Makefile, implies building onnx
  - [x] simply add the generated files and update when onnx submodule is updated
    - my choice [alrevuelta: agree. Where do you plan to add the call to the python script?]
- how to handle the different types?
  - [x] convert current stub into switch for type specific implementations and stub these type implementations [alrevuelta: My choice also. As a random idea, would be nice to have the possibility of not including all the types in the binary. I.e. if I want to run inference on a model that is using only the double implementation, maybe there is a way to "skip" the other types and have a smaller binary, which might be useful in some constrained devices. Just an idea though.]
    - allows partial implementations
    - stub can check if right types were supplied
    - how to resolve different types between inputs?
    - if possible, my choice
  - single implementation handles all types
- should the operator return value be a custom enum? [alrevuelta: Will be scarify a lot of performance? Your suggestion seems fine to me (ENOSYS, ENOMEM) are important, but I don't really understand EDOM or ERANGE. If the onnx is correct we should never run into overflow issues, right?]
  - [x] yes, we have cases errno is not sufficient
  - no, we can do everything with errno
    - documents which return values are expected
- should we respect onnx domains?
  - currently 3 existing domains: `''`, `ai.onnx.ml'`, `ai.onnx.training'`
  - may produce name conflicts in the future
  - [x] rename operators to reflect domains
    - my choice [alrevuelta: Yep, totally agree. I would both rename and restructure dirs]
  - [x] restructure dirs to reflect domains
    - would allow to generate domain libraries, my choice
  - [x] rename files to reflect domains
    - not necessary if we split up dirs
- should we integrate 'old' operators?
  - onnx redefined some operators
  - produces naming conflicts
  - what if our implementations become 'old'?
  - [x] rename operators to reflect version they were introduced
    - my choice [alrevuelta: Agree. Lets make this project as flexible and maintainable as possible]


nopeslide:
> - how to manage generated files?
>   - [ ] should be generated with main Makefile, implies building onnx
>   - [x] simply add the generated files and update when onnx submodule is updated
>     - my choice [alrevuelta: agree. Where do you plan to add the call to the python script?]
> 
@alrevuelta would integrate the onnx build and generation into the main Makefile as an optional target

> - how to handle the different types?
>   - [x] convert current stub into switch for type specific implementations and stub these type implementations [alrevuelta: My choice also. As a random idea, would be nice to have the possibility of not including all the types in the binary. I.e. if I want to run inference on a model that is using only the double implementation, maybe there is a way to "skip" the other types and have a smaller binary, which might be useful in some constrained devices. Just an idea though.]
>     - allows partial implementations
>     - stub can check if right types were supplied
>     - how to resolve different types between inputs?
>     - if possible, my choice
>   - [ ] single implementation handles all types
>
@alrevuelta good idea. I would do this via the linker by generating a library per domain and type
So we have the stub inside the binary, but no actual implementation. The stub size can be reduced to a simple `return ENOSYS`. 

> - should the operator return value be a custom enum? [alrevuelta: Will be scarify a lot of performance? Your suggestion seems fine to me (ENOSYS, ENOMEM) are important, but I don't really understand EDOM or ERANGE. If the onnx is correct we should never run into overflow issues, right?]
>   - [ ] yes, we have cases errno is not sufficient
>   - [ ] no, we can do everything with errno
>
@alrevuelta I included these errors just to show what errors we could expect from `math.h`
An enum would cost nothing (both resolve to simple ints), but would allow us to specify an operator interface (only these codes may be produced and should be handled)

nopeslide:
> @alrevuelta good idea. I would do this via the linker by generating a library per domain and type
> So we have the stub inside the binary, but no actual implementation. The stub size can be reduced to a simple `return ENOSYS`.
>
I could also introduce a global stub and redirect the linker to this single stub if the symbol isn't defined. This may even solve the problem of having a stub for each input type permutation.

alrevuelta:
@nopeslide I like the idea of adding `onnx` as a submodule but maybe we can do it in a simpler way. What if we specify the specific official release(s) that we support (i.e. [`v1.6.0`](https://github.com/onnx/onnx/releases/tag/v1.6.0)) rather that a specific commit?

If we do it like that, we can avoid having the `onnx` makefile and building it. We can just use the power of Python and pip. So with pip we install whatever version we are using and then `C.defs.get_all_schemas()` will have all the data for that release. This version information could be for example in a `requirements.txt`.

Maybe its not needed but doing it this way it will be very simple to add it to the CI. One idea aside could be to autogenerate in CI all the operators status that [you made](https://github.com/alrevuelta/cONNXr/blob/master/doc/06_OperatorStatus.md)

nopeslide:
> @nopeslide I like the idea of adding `onnx` as a submodule but maybe we can do it in a simpler way. What if we specify the specific official release(s) that we support (i.e. [`v1.6.0`](https://github.com/onnx/onnx/releases/tag/v1.6.0)) rather that a specific commit?
> 
> If we do it like that, we can avoid having the `onnx` makefile and building it. We can just use the power of Python and pip. So with pip we install whatever version we are using and then `C.defs.get_all_schemas()` will have all the data for that release. This version information could be for example in a `requirements.txt`.
> 
@alrevuelta this strips the option of being bleeding edge and moves the build process to the distributor, what may get us into problems regarding build options. Who guarantees the provided library is the one we need?

> Maybe its not needed but doing it this way it will be very simple to add it to the CI. One idea aside could be to autogenerate in CI all the operators status that [you made](https://github.com/alrevuelta/cONNXr/blob/master/doc/06_OperatorStatus.md)
> 
@alrevuelta this information should be an easy query of the binary with `nm` and would be the next step ;D

> Couple of things more:
> 
>     * Lets say that we have few operators implemented. If you run the python script as it is it will regenerate all c/h files and destroy the already implemented ones.
> 
@alrevuelta I'm rewriting the script to only generate headers (only possible with the global stub). So no implementation will be overwritten. The actual implementations should not even need the generated header, since the linker handles the replacement.
 
>     * Just curious. Any reason behind having one `.h` file per operator? Can't we have a huge common header file for all operators?
@alrevuelta mostly debugging reasons, so I can generate a single headers and see simple diffs. we could generate a massive header file (up to 100k loc), but this would also make it harder to see changes when we switch onnx versions.

alrevuelta:
> @alrevuelta this strips the option of being bleeding edge and moves the build process to the distributor, what may get us into problems regarding build options. Who guarantees the provided library is the one we need?

@nopeslide Which kind of problems can we get regarding the build options? _Who guarantees the provided library is the one we need?_ well if I'm not wrong and we set a dependancy to a specific onnx version in the `requirements.txt` (which btw is recognised by GitHub under "Insights" -> "Dependency graph") thats the guarantee that the library we are providing is the one we need. When calling `C.defs.get_all_schemas()` we should be calling the exact same code as the one you are calling. Quick example:

Lets say that we are working with commit `553df22c67bee5f0fe6599cff60f1afc6748c635`. With your solution you would have under `third_party/onnx` the folder `onnx @553...`.
That commit, corresponds to the latest release `v1.6.0`, so with my solution we would have the `requirements.txt` file with the content `onnx==1.6.0`, which will tell that whatever python script is run in our repo, needs to have that onnx version. Both runs should give the same results, with the difference that we target a release rather than a commit (which looks cleaner for me).

Just trying here to avoid compiling what has already been compiled by someone else. No need to clone the whole repo and easier to install in whatever platform with `pip`.

I'm also insisting on this because it might be useful when testing. There is a script in `onnx.backend.test.cmd_tools` called data `generate_data` that generates all the tests for all operators (the `.onnx` + `.pb`) that we have copy-pasted in our repo. Was your testing idea related to this?

nopeslide:
> > @alrevuelta this strips the option of being bleeding edge and moves the build process to the distributor, what may get us into problems regarding build options. Who guarantees the provided library is the one we need?
> 
> @nopeslide Which kind of problems can we get regarding the build options? _Who guarantees the provided library is the one we need?_ well if I'm not wrong and we set a dependancy to a specific onnx version in the `requirements.txt` (which btw is recognised by GitHub under "Insights" -> "Dependency graph") thats the guarantee that the library we are providing is the one we need. When calling `C.defs.get_all_schemas()` we should be calling the exact same code as the one you are calling. Quick example:
> 
> Lets say that we are working with commit `553df22c67bee5f0fe6599cff60f1afc6748c635`. With your solution you would have under `third_party/onnx` the folder `onnx @553...`.
> That commit, corresponds to the latest release `v1.6.0`, so with my solution we would have the `requirements.txt` file with the content `onnx==1.6.0`, which will tell that whatever python script is run in our repo, needs to have that onnx version. Both runs should give the same results, with the difference that we target a release rather than a commit (which looks cleaner for me).
> 
> Just trying here to avoid compiling what has already been compiled by someone else. No need to clone the whole repo and easier to install in whatever platform with `pip`.
> 
@alrevuelta pip itself builds the source each time you install the onnx package (at least on my platform). Further, the current version "1.6.0"  (what pip uses) is broken with the latest protobuf version. A fix exists upstream, the submodule tracks it, but I can't build onnx via pip on my platform since I have the latest protobuf version. 

> I'm also insisting on this because it might be useful when testing. There is a script in `onnx.backend.test.cmd_tools` called data `generate_data` that generates all the tests for all operators (the `.onnx` + `.pb`) that we have copy-pasted in our repo. Was your testing idea related to this?
@alrevuelta Would be the next thing, I limited the onnx build just to the things I need for the stub generation.

nopeslide:
> @alrevuelta I'm rewriting the script to only generate headers (only possible with the global stub). So no implementation will be overwritten. The actual implementations should not even need the generated header, since the linker handles the replacement.

@alrevuelta 
how do we want to implement the choice of the right implementation (regarding type)?
Currently I'm writing the type-checker/type-switch of the generator script.
Should these checks run every time the operator is called? (kind of slow I assume)
Or should I write a "resolve" function which returns the right implementation for the specified input?
As long the input type doesn't change (which shouldn't, should it?) we could reuse the inital resolved function.

alrevuelta:
> @alrevuelta pip itself builds the source each time you install the onnx package (at least on my platform). Further, the current version "1.6.0" (what pip uses) is broken with the latest protobuf version. A fix exists upstream, the submodule tracks it, but I can't build onnx via pip on my platform since I have the latest protobuf version.

@nopeslide  No sure what pip does under the hood but for me it takes just some seconds to install, more than cloning the repo and compiling.

> how do we want to implement the choice of the right implementation (regarding type)?
Currently I'm writing the type-checker/type-switch of the generator script.
Should these checks run every time the operator is called? (kind of slow I assume)
Or should I write a "resolve" function which returns the right implementation for the specified input?
As long the input type doesn't change (which shouldn't, should it?) we do not need to resolve the type again

@nopeslide Yep, thats something that I don't really know how to solve, I guess a switch is the way to go. The input type **can change** along the model (see operator Cast). An example would be a model that uses fixed point representation, where the input is lets say double and the calculations are made in int16 and the output is returned again as a double.

scripts/onnx_generator/OnnxWrapper.py
```diff
@@ -0,0 +1,490 @@
...
+        self.ref_doc = None
+        self.range_input = None
+        self.range_output = None
+        self.ref_file = None
+        if isinstance(schema, dict):
+            self.__dict__.update(schema)
+        else:
+            self.doc = OnnxDoc(schema)
+            self.name = schema.name
+            self.deprecated = schema.deprecated
```
alrevuelta, line 457: Just FYI, `deprecated` doesn't exist in some older versions (i.e. 1.0.0)

scripts/onnx_generator/__main__.py
```diff
@@ -0,0 +1,87 @@
+import argparse
+import os
+import sys
+import re
+from .OperatorHeader import OperatorHeader
```
alrevuelta, line 5: I'm having issues both in the terminal and pycharm with the leading `.`. Can we just remove it?

scripts/onnx_generator/__main__.py
```diff
@@ -0,0 +1,87 @@
...
+      schemas.remove(s)
+  print(f"pattern '{pattern}' excluded {len(excluded)} operators")
+
+headers = [ OperatorHeader(s,args.header[0]) for s in schemas ]
+
+for h in headers:
+  filename = f"{h.schema.operator_name}.h"
+  dirname = f"{args.path[0]}/{args.header[0]}/{h.schema.domain}/"
+  if args.verbose:
+    print(f"writing header {dirname}/{filename}")
```
alrevuelta, line 84: Extra `/`? Looks like this is printing `include/operators/onnx**//**operator__onnx__gemm__1.h`

third_party/onnx/Makefile
```diff
@@ -0,0 +1,38 @@
+
```
alrevuelta, line 1: Just realised that we have another extra dependancy from `onnx` repo, which is the `onnx.proto` file. The files `onnx.pb.c` and `.h` that we have are generated based on that `.proto` file. Looks like this might be a good place to generate that files. I was doing it with `protoc` library:
```
protoc --c_out=. onnx.proto
```

scripts/generate_onnx_stubs.py
```diff
@@ -0,0 +1,695 @@
+import inspect
+import itertools
+import os
+import sys
+import re
+
```
alrevuelta, line 6: Is this file still being used?

scripts/onnx_generator/__main__.py
nopeslide: should be no problem, multiple slashes are ignored by the os (at least on *nix systems).
but could be passed through the python path lib again

scripts/generate_onnx_stubs.py
nopeslide: I try to replace this draft with the module. will delete it when I'm sure I didn't miss anything

nopeslide:
> Had a quick look, looks neat! Added my 5 cents and looking forward to see it in action.
> 
> Quick question regarding `resolve function for operator type` bullet. Where do you plan to do this? In `.c` files? I'm trying to understand where do you plan to resolve the type. Are you going to generate one `.c` file for each operator where the correct function type is called?

asked me the same thing, the first idea was exactly what you assumed:
Generate a resolver which returns the correct function pointer and generate a typeless operator function which always calls the resolver and executes the returned function.
So we have the option of just using the typeless operator or "cache" the resolved functions.  
If you have a better proposal, let me hear it.

scripts/onnx_generator/__main__.py
nopeslide: I can't execute it without it. did you run it as a model (`python -m onnx_generator`)?

scripts/onnx_generator/OnnxWrapper.py
nopeslide: good to know. thanks

third_party/onnx/Makefile
nopeslide: does this file also come with the onnx pip installation? just asking to determine if it makes sense to integrate a switch in the makefile to use the onnx repo or the installed version

nopeslide:
@alrevuelta please have look a the current state.
keep in mind that the generated code was never put through gcc: ignore syntax errors, just assume it's pseudo code :)

I'm not quite satisfied that the resolve function does all the sanity checking.
How about I split this functionality up into 2 different function? one resolving the the constraints and the other checking if the operator can work on the supplied parameters?

Can I assume this onnx statement holds true for connxr?
`Input and outputs are positionally associated with operator inputs and outputs. Attributes are associated with operator attributes by name.` [source](https://github.com/onnx/onnx/blob/master/docs/IR.md)
If we take a look at the [batchnormalization operator](https://github.com/onnx/onnx/blob/master/docs/Operators.md#batchnormalization). this operator has a lot of optional outputs. does this mean if we want to get the second optional `output_var` output we have to specify a dummy for the first optional `output_mean` output or is just the order fixed (`output_mean` is skipped). I'm asking because the lookup of names is quite time-consuming and if we skip those optional inputs/outputs we always have to find the correct tensor.

third_party/onnx/Makefile
alrevuelta: Yes, assuming you have `onnx` package installed, you can access the file with:

```python
import pkg_resources
path = "onnx.proto"
filepath = pkg_resources.resource_filename("onnx", path)
print(filepath)
```

alrevuelta:
@nopeslide Looks really nice, thanks a lot for this great work!

> I'm not quite satisfied that the resolve function does all the sanity checking.
How about I split this functionality up into 2 different function? one resolving the constraints and the other checking if the operator can work on the supplied parameters?

I like having all these sanity checks, but sometimes I have second thoughts and think its too much. Shouldn't our code be more dummy? Can't we assume that the model is well defined and all the number of inputs/outputs and types are correct? Or maybe we can just run all the checks when the model is loaded and not on runtime? onnx provides `onnx.checker.check_model(onnx_model)`, but haven't checked what it does. Lets say we are running inference in a real world scenario, some times per second. Do we really need to check that the inputs/outputs and types are correct for every inference?

But looking at what we have, I can see three types of checks:
* 1) Input and outputs sizes
* 2) Same tensor type
* 3) Tensor type

Isn't `2)` contained in `3)`?. If all tensors have the correct type, that would imply `2`, wouldn't it? 

Any reason why you are not checking also the number of attributes?

> Can I assume this onnx statement holds true for connxr?
Input and outputs are positionally associated with operator inputs and outputs. Attributes are associated with operator attributes by name. source

Really good point. Currently I'm just searching for the tensor by its name but I agree its very time consuming. Will investigate further on this and give you an answer.

nopeslide:
>I like having all these sanity checks, but sometimes I have second thoughts and think its too much. Shouldn't our code be more dummy? 

I completely concur with you, hence the question if I should separate check and resolve functionalities so we can run an optional check of the network at startup and then resolve each operator at will without the overhead. I would also split these two functions into different files, so we can easily select what tests to compile. I just included the testing, because I thought we could use it while debugging stuff.
The onnx wrapper is also capable of introducing new operator schemas without touching the onnx framework, therefore the checks make sense because onnx can't verify these extensions.

> Isn't 2) contained in 3)?. If all tensors have the correct type, that would imply 2, wouldn't it?

in 2) I just check if they share the same type, no type validation happens
in 3) I check each input and output if they have a valid type 
currently 3) checks all tensors, regardless of shared types, this is of course overkill if they share a type, but I liked the simplicity of non-dependent checks.

> Any reason why you are not checking also the number of attributes?

Didn't come around for it, would be the next thing. I also ignore the optional flag of inputs/outputs at the moment.

nopeslide:
@alrevuelta 
> Really good point. Currently I'm just searching for the tensor by its name but I agree its very time consuming. Will investigate further on this and give you an answer.

How about I generate a context struct for each operator? this would also simplify the actual operator implementation. sth like:
```c
struct operator__onnx__unique__11_context {
  struct {
    Onnx__TensorProto *X;
  } in;
  struct {
    Onnx__TensorProto *Y;
    Onnx__TensorProto *indices;
    Onnx__TensorProto *inverse_indices;
    Onnx__TensorProto *counts;
  } out;
  struct {
    int axis;
    int sorted;
  } attr;
  onnx_operator run;
};
```
the resolve function could simply fill such a struct and specify the callback (`run` in this example).

alrevuelta:
> I completely concur with you, hence the question if I should separate check and resolve functionalities so we can run an optional check of the network at startup and then resolve each operator at will without the overhead. I would also split these two functions into different files, so we can easily select what tests to compile. I just included the testing, because I thought we could use it while debugging stuff.

Yes, lets do that. Lets separate the checks from the resolver and run it at startup.

> How about I generate a context struct for each operator? this would also simplify the actual operator implementation. sth like:

In that case we would have to define one context struct per operator, maybe to much. And I also liked the current implementation, where all operators share the same interface. With this approach each operator would have a different input type.
What about having a common one where optional fields are NULL and the order follows the spec? Using BatchNormalization as an example:
* Inputs: X, scale, B, mean, var, training_mode(optional)
*Outputs: Y, output_mean(optional), output_var(optional), saved_mean(optional), saved_var(optional)
* Attributes: momentum(optional), output_mean(optional)

```c
struct operator__context {
    Onnx__TensorProto **inputs;
    Onnx__AttributeProto **attributes;
    Onnx__TensorProto **outputs;
    onnx_operator run;
};
```
That structures will be filled at startup, so we don't have to search anymore and `inputs` will always have size 6, `outputs` 5, and `attributes` 2. With this we wouldn't need `n_input`, `n_attribute` and `n_output`.

And I really like this idea "the resolve function could simply fill such a struct and specify the callback (run in this example)." If I have correctly understood, you want to resolve the function at startup son on run time we already know which function to call.

nopeslide:
> And I really like this idea "the resolve function could simply fill such a struct and specify the callback (run in this example)." If I have correctly understood, you want to resolve the function at startup son on run time we already know which function to call.

yes

> In that case we would have to define one context struct per operator, maybe to much. And I also liked the current implementation, where all operators share the same interface. With this approach each operator would have a different input type.

how about we combine both approaches? :D 
the operator context will be kept generic like you proposed:
```c
struct operator__context {
    Onnx__TensorProto **inputs;
    Onnx__AttributeProto **attributes;
    Onnx__TensorProto **outputs;
    onnx_operator run;
};
```
the tensors in the list have a fixed position and are NULL if not specified. 
Further I generate these custom structs for each operator:
```c
  struct {
    Onnx__TensorProto *X;
  } operator__onnx__unique__11_input;

  struct {
    Onnx__TensorProto *Y;
    Onnx__TensorProto *indices;
    Onnx__TensorProto *inverse_indices;
    Onnx__TensorProto *counts;
  } operator__onnx__unique__11_output;

  struct {
    Onnx__AttributeProto *axis;
    Onnx__AttributeProto *sorted;
  } operator__onnx__unique__11_attribute;

struct {
  struct operator__onnx__unique__11_input *in;
  struct operator__onnx__unique__11_output *out;
  struct operator__onnx__unique__11_attribute *attr;
  onnx_operator run;
} operator__onnx__unique__11_context;
```

inside the implementation we can simply cast the generic context to the specific one:
```c
int operator__onnx__unique__11(struct operator__context *generic_context) {
  struct operator__onnx__unique__11_context *specific_context = (void *) context;

// access input tensor 'X'
  specific_context->in->X // same as generic_context->inputs[0]
...
}
```

but what about variadic tensors?
we could NULL-terminate each pointer list to accommodate for variadic tensors,
but wouldn't it be practical to directly know the number of variadic arguments?

alrevuelta:
@nopeslide Looks good to me. And being able to do `specific_context->in->X` is quite convenient and improves readability.

> but what about variadic tensors?
we could NULL-terminate each pointer list to accommodate for variadic tensors,
but wouldn't it be practical to directly know the number of variadic arguments?

Can you elaborate more on the variadic function approach? How is the function prototype if we have two different types, tensor and attribute. And how are the inputs handled within the function?

nopeslide:
@alrevuelta 
> Can you elaborate more on the variadic function approach? How is the function prototype if we have two different types, tensor and attribute. And how are the inputs handled within the function?

I meant variadic tensors featured by onnx (i.e. see [Adagrad](https://github.com/onnx/onnx/blob/master/docs/Operators.md#aionnxtraining))
We could just append all of these to the tensor pointer list and define them as variable array inside the specific struct:
```c
struct operator__ai_onnx_training__adagrad__1_input {
  Onnx__TensorProto *R;
  Onnx__TensorProto *T;
  Onnx__TensorProto *inputs[];
};
```
by NULL-terminating the pointer list we could enable simple iteration over them, but I assume there are cases in which the number of tensors may be a nice information to have beforehand.

alrevuelta:
@nopeslide I see, so your suggestion is to combine the variadic tensor with the solution we already discussed? I like the idea, and its also a great way of supporting operators with an undefined number of i/o.

NULL-terminating the sequence seems good to me, but I agree that having the `n_of_tensors` might be good in some occasions. Maybe adding it is too much, so I would go with the NULL-terminated way.

nopeslide:
> Maybe adding it is too much, so I would go with the NULL-terminated way.

too much in what way?
memory-wise a null termination costs the same as a counter.
execution-wise there should be no difference in calling since we pass the context struct as parameter.

alrevuelta:
@nopeslide Right. In that case which value do you plan to store? The total number of input/outputs? Or only the number of non-fixed i/o? I would go with the second.

nopeslide:
have a look at [loop](https://github.com/onnx/onnx/blob/master/docs/Operators.md#loop).
it features optional inputs and an additional variadic input.
onnx seems to solve the problem by forcing the user to insert default tensors for the optional ones.
I think we may overengineer this one. let us just do the NULL-termination and see if we get into any problems. the struct can always be extended without breaking anything.

nopeslide:
> However I can't come up with a good solution for the outputs. Some nodes inputs depend on previous nodes outputs to work. So some nodes inputs, will point some other nodes outputs. And if we do that at startup, it will point to a memory position that is empty at that moment, which is not problem ofc.
> My point here is that for populating all inputs for a given operator, we need some knowledge on past operators, and this is something that I don't think is trivial.
> Any ideas?

If I understand you correctly, you are worried about the initialization order?
If so, isn't this the same problem we are also facing when scheduling nodes with multiple threads?
To schedule multiple threads we would need a dependency graph telling us how the nodes depend on each other and which nodes may be executed in parallel.
And this dependency graph should also tell us in what order the need to be initialized.

alrevuelta:
> If I understand you correctly, you are worried about the initialization order?
If so, isn't this the same problem we are also facing when scheduling nodes with multiple threads?
To schedule multiple threads we would need a dependency graph telling us how the nodes depend on each other and which nodes may be executed in parallel.
And this dependency graph could also tell us in what order the need to be initialized.

@nopeslide 
Not exactly, so far we are assuming that once we reach a node, all its inputs have been previously computed. So that should be ok for now.

I will try to explain it with a quick patch, that we can also use for prototyping and discussions.

alrevuelta:
@nopeslide Created a quick PR #11 to discuss the implementation details. I think its better to have it working with some simple model like MNIST and everything that Python generates hardcoded, and then we can scale it up to Python and this PR. The patch is just pseudocode.

nopeslide:
sorry I missed some thinks

nopeslide:
@alrevuelta 
It compiles and throws no warnings or errors for the minimal mnist generation. So we could try to use the generated code for #11

alrevuelta:
@nopeslide Sorry for such a late reply.

> variadic support in the context generation
via variable array? (does only support one variadic tensor in input and output)
via pointer to second struct?
resolve function filling the context (makes no sense to generate this, because it's runtime code)

So far I don't have an opinion regarding variadic support.
Regarding the function filling the context, I assumed it was going to be autogenerated code. Are you saying we have to write the code to populate the structures that contain the context? That implies searching for the tensors, and having specific knowledge of i/o/attributes name. Either way, lets not add more to this PR.

> It compiles and throws no warnings or errors for the minimal mnist generation. So we could try to use the generated code for #11 (just run make onnx_generator (with ONNX_CUSTOM= if you want to use pip onnx))

Yes, thats a great idea. How should we proceed?, my idea:
* Generate all the files in this commit and merge it to master, since its not breaking anything.
* In #11 use the context and start migrating the old operators to the new interface.
* And the rest we will see, because as you said, this PR is huge and its becoming difficult to see whats going on.

nopeslide:
> So far I don't have an opinion regarding variadic support.

lets make the varaidic problem an issue so we have a reminder when we come around implementing operators with variadic tensors.

> Regarding the function filling the context, I assumed it was going to be autogenerated code. Are you saying we have to write the code to populate the structures that contain the context? That implies searching for the tensors, and having specific knowledge of i/o/attributes name. Either way, lets not add more to this PR.

what I meant was: filling and resolving these structs is a generic algorithm. it makes no sense to generate a custom one for each operator.

the current algorithm (I had in mind) would look like this:
1. get onnx version from pb.
2. find matching operand set in generated set structure
3. iterate through nodes in pb
   1. find resolve function for  each node in set (we should save the resolver in the context)
   2. malloc/construct attribute, input and output list and fill with names from pb
   3. set resolved tensor if it is an initializer (no need for each operator to resolve from pb)
4. iterate through input list of created contexts
  1. if tensor is unresolved, find and set originating context
5. resolve output generating context
   1. find all input tensors
       1. if input tensor is unset, resolve and execute originating context
       2. search in originating context output array the correct tensor
   2. execute

this step no. 5 would be a recursive call on the output node, walking up to the networks input and executing them one after another on its way back (depth-first execution).

alrevuelta:
@nopeslide 

> what I meant was: filling and resolving these structs is a generic algorithm. it makes no sense to generate a custom one for each operator.

I don't think it is that generic. Populating the context will need some knowledge of the specific one. For example, if we are populating the attribute context of `conv`:

```c
typedef struct operator_context_attribute__operator__onnx__conv__11 {
    size_t length;
    Onnx__AttributeProto *auto_pad;
Onnx__AttributeProto *dilations;
Onnx__AttributeProto *group;
Onnx__AttributeProto *kernel_shape;
Onnx__AttributeProto *pads;
Onnx__AttributeProto *strides;
} operator_context_attribute_operator__onnx__conv__11
```
 Lets say we are working with a `conv` operator with `auto_pad`, `kernel_shape` and `strides`. So we would have in the attribute list for that node three attributes with those names. We can get the pointer to each on of them, but how do we fill the structure? Some of them can be missing (if they are optional) and without specific code for each structure I don't know how we can handle those gaps.

Its more or less what I was doing here: https://github.com/alrevuelta/cONNXr/pull/11/files#diff-dbd4019d3365b93edeaa89f8607d6d45R75

Makefile
```diff
@@ -187,4 +218,33 @@ clean_profiling:
...
+onnx_generator:
+	cd scripts; python -m onnx_generator \
+	$(if $(ONNX_CUSTOM), --onnx $(abspath $(ONNX_CUSTOM))) \
+	$(if $(ONNX_INCLUDE), --include $(ONNX_INCLUDE)) \
+	$(if $(ONNX_EXCLUDE), --exclude $(ONNX_EXCLUDE)) \
+	$(if $(ONNX_VERSION), --version $(ONNX_VERSION)) \
+	$(if $(ONNX_DOMAINS), --domains $(ONNX_DOMAINS)) \
+	-vv \
+	--force \
+	$(shell git rev-parse --show-toplevel)
```
alrevuelta, line 248: Not working on macOS. Is there a way we can make it compatible?

nopeslide: @alrevuelta 
is this still relevant?

nopeslide:
> @nopeslide
> 
> > what I meant was: filling and resolving these structs is a generic algorithm. it makes no sense to generate a custom one for each operator.
> 
> I don't think it is that generic. Populating the context will need some knowledge of the specific one. For example, if we are populating the attribute context of `conv`:
> 
> ```c
> typedef struct operator_context_attribute__operator__onnx__conv__11 {
>     size_t length;
>     Onnx__AttributeProto *auto_pad;
> Onnx__AttributeProto *dilations;
> Onnx__AttributeProto *group;
> Onnx__AttributeProto *kernel_shape;
> Onnx__AttributeProto *pads;
> Onnx__AttributeProto *strides;
> } operator_context_attribute_operator__onnx__conv__11
> ```
> 
> Lets say we are working with a `conv` operator with `auto_pad`, `kernel_shape` and `strides`. So we would have in the attribute list for that node three attributes with those names. We can get the pointer to each on of them, but how do we fill the structure? Some of them can be missing (if they are optional) and without specific code for each structure I don't know how we can handle those gaps.
> 
> Its more or less what I was doing here: https://github.com/alrevuelta/cONNXr/pull/11/files#diff-dbd4019d3365b93edeaa89f8607d6d45R75

You are absolutely right

include/operators/operator_stub.h
```diff
@@ -0,0 +1,15 @@
+#ifndef OPERATOR_STUB_H
```
alrevuelta, line 1: Just realised. Is this some legacy leftover?

alrevuelta: @nopeslide

nopeslide: It is currently returned by the type resolver if no matching operator is found

alrevuelta: I expected the stub to have the same inputs as the operators, like the generic `operator_context`.

```c
struct operator_context
{
    operator_context_input      *input;
    operator_context_output     *output;
    operator_context_attribute  *attribute;
    operator_executer operator;
};
```

nopeslide: thanks! didn't notice. fixed it in #15

Makefile
nopeslide: I'm asking because you merged

PR #11, opened 2020-04-22
Prototype for #10

alrevuelta:
@nopeslide 
Prototype for #10, just some pseudocode to discuss about the implementation.

### DIFF
--- a/src/inference.c
+++ b/src/inference.c
@@ -9,35 +9,12 @@
 int _outputIdx = 0;
 Onnx__TensorProto *_outputs[MAX_NUM_OF_OUTPUTS] = {};
 
-static int call_operator(char *name,
-                         size_t n_input,
-                         Onnx__TensorProto **input,
-                         size_t n_attribute,
-                         Onnx__AttributeProto **attribute,
-                         size_t n_output,
-                         Onnx__TensorProto **output)
-{
-  int i;
-  for(i = 0; i < NUMBER_OF_OPERATORS; i++) {
-    if(strcmp(operatorsSet[i].name, name) == 0) {
-      int ret = operatorsSet[i].func(n_input,
-                                     input,
-                                     n_attribute,
-                                     attribute,
-                                     n_output,
-                                     output);
-      return ret;
-    }
-  }
-  /* If it reaches this point, the operator wasnt found. Throw error? */
-  TRACE_LEVEL0("\n\nTODO: Operator %s doest not exist or its not implemented\n\n", name);
+Onnx__TensorProto** lazy_outputs_mapping_tensors[MY_TABLE_SIZE] = {};
 
-  /* break */
-  perror("There was an error calling the operator");
-  exit(1);
+char lazy_output_mapping_names[MY_TABLE_SIZE][MAX_STRING_SIZE];
 
-  return -1;
-}
+operator__context** all_op_context;
+
 
 Onnx__TensorProto** inference(Onnx__ModelProto *model, Onnx__TensorProto **inputs, int nInputs)
 {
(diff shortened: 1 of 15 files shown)


nopeslide:
I think I got it.
you not worried about how to resolve dependencies,
but how to resolve the tensors, since they are named arbitrarily by the network and not by the operator using them?

src/inference.c
```diff
@@ -103,3 +119,79 @@ Onnx__TensorProto** inference(Onnx__ModelProto *model, Onnx__TensorProto **input
...
+      names depending on the node:
+      -Convolution28_Output_0
+      -Convolution110_Output_0
+      -Times212_Output_0
+      */
+
+      /* This wont work as it is. For a given node, their inputs come from
+      three different sources.
+      1)Existing initializers in the model: model->graph->initializer[i]->name
+      2)Provided input to the model: inputs[i]->name
```
nopeslide, line 154: isn't this the same thing? all provided initializers are registered as input

alrevuelta:
@nopeslide Yes, exactly. AFAIK there are 3 types of input tensors in a node:
* The initializer ones. Not a problem, they are part of the model and we can search them by their name.
* The input ones. These are the input tensors to the model (i.e. in MNIST is only one and its the image). Also not a problem, we can search them by name.
* And the inputs that are some other nodes output. So far we are not storing the name to pointer mapping. So for example if the output of an operator is `O`, we would have it stored in the context `operator__onnx__xxx__output` and we can of course access it `->O` but some other node will need that output as input, but that node doesn't know about `O`. It only knows about a name i.e. `Output_From_Add_Operator`. So we need to think about storing this mapping somewhere.

src/inference.c
alrevuelta: I guess we can say its the same thing, but just wanted to point out that they come from different sources.

nopeslide:
> So we need to think about storing this mapping somewhere.

Instead referencing the `Onnx__TensorProto` objects directly we could make them resolvable:
```c
struct operator_tensor {
  char                    *name;   // name given by model
  struct operator_context *ctx;    // context the tensor originated from
  Onnx__TensorProto       *tensor; // resolved tensor
};

struct operator_context{
  struct operator_tensor   **input;
  size_t                     n_input;

  struct operator_tensor   **output;
  size_t                     n_output;

  struct operator_attribute *attr;
  size_t                     n_attr;

  onnx_operator run;
  int status; //error code
};
```
The list of input and output tensors can still be overlayed by the specific context of an operator and we can still replace unspecified tensors with a NULL pointer.
Additionally we now have knowledge about the dependencies of our current node and can use this to create a execution schedule.

nopeslide:
Btw I played around with on-demand resolving of inputs and found the length specified (`n_*`) pretty handy. would prefer them now over NULL-termination

alrevuelta:
@nopeslide Is it possible to have the function pointer inside the `operator_context` and then the `operator` context as part of the function pointer? Isn't this some kind of circular dependancy. I can't get it to work.

```c
struct operator_context{
  ... 
  onnx_operator run;
 ...
};
```

```c
typedef int (*onnx_operator)(
  operator__context *context
);
```

nopeslide:
> @nopeslide Is it possible to have the function pointer inside the `operator_context` and then the `operator` context as part of the function pointer? Isn't this some kind of circular dependancy. I can't get it to work.
> 
> ```c
> struct operator_context{
>   ... 
>   onnx_operator run;
>  ...
> };
> ```
> 
> ```c
> typedef int (*onnx_operator)(
>   operator__context *context
> );
> ```

typedefs are definitions and not declarations. to solve such circular declarations you may need forward declarations, which typedefs can't provide:
```c
struct A;
struct B;

typedef struct A {
  struct B *b;
} A;

typedef struct B {
  struct A *a;
} B;
```
you also can only use pointers on forward declarations (since the struct can't be filled)

alrevuelta:
@nopeslide Just uploaded a new patch with some more nasty code. Now I think we can play around with it and come up with a final design solution. Iterates all nodes and calls the previously resolver operator with the previously resolved context.

MNIST operator is working and now the inference function is just:

```c
  for (int nodeIdx = 0; nodeIdx < n_nodes; nodeIdx++)
  {
    all_op_context[nodeIdx]->operator(all_op_context[nodeIdx]);
  }
```

The code that would be generated is under the function `resolve_check_get_input_and_attr`. It gets the model with its inputs, and returns an array of contexts, with the operator, i/o/attr resolved ready to be used to run inference.

The context for each operator looks like this:
```
struct operator__onnx__maxpool__context{
  struct operator__onnx__maxpool__input *in;
  struct operator__onnx__maxpool__output *out;
  struct operator__onnx__maxpool__attribute *attr;
  int (*operator)(struct operator__context*);
};
```

```c
struct operator__onnx__maxpool__input{
  Onnx__TensorProto *X;
};
```

Just measured the times of the inference (excluding the resolving process ofc) for mnist and looks a bit higher. Don't know why, now that we have everything resolved and no time is wasted on searching I expected it to be quicker. The only difference in the operators is the cast:
```c
struct operator__onnx__matmul__context *sc = (void *) context;
```

Will continue looking at it, but I think we need to start taking the decision on how the context is going to look like.

alrevuelta:
> Instead referencing the `Onnx__TensorProto` objects directly we could make them resolvable:
> ```c
> struct operator_tensor {
>   char                    *name;   // name given by model
>   struct operator_context *ctx;    // context the tensor originated from
>   Onnx__TensorProto       *tensor; // resolved tensor
> };
> 
> struct operator_context{
>   struct operator_tensor   **input;
>   size_t                     n_input;
> 
>   struct operator_tensor   **output;
>   size_t                     n_output;
> 
>   struct operator_attribute *attr;
>   size_t                     n_attr;
> 
>   onnx_operator run;
>   int status; //error code
> };
> 
> Onnx__TensorProto*
> resolve_tensor(operator_tensor *tensor) {
>   if (!tensor->tensor) {
>     tensor->tensor = findOutputTensorInContext(tensor->name, tensor->ctx);
>   }
>   return tensor->tensor;
> }
> ```
> The list of input and output tensors can still be overlayed by the specific context of an operator and we can still replace unspecified tensors with a NULL pointer.
> Additionally we now have knowledge about the dependencies of our current node and can use this to create a execution schedule.

I like the simplicity of this solution, which is similar to what we already had on master. We would still need to search the tensor but only on operator level, among less entries. I have played a bit with our context specific idea, and being able to address the tensor like in->X is really convenient, but it adds some complexity on the Python side. Lots of structures, and quite a lot of autogenerated code for input/output/attributes resolving (which so far is the only drawback I have seen). Even though, I like this solution and I would move forward with this. The Python autogenerated code was thought to eliminate boiler plate code, and with this we could avoid some lines per operator.
But maybe I'm missing something here and you have other thoughts. Lets discuss it and close this so we can move forward :)

nopeslide:
> But maybe I'm missing something here and you have other thoughts. Lets discuss it and close this so we can move forward :)

Autogenerating these should be no problem, will do.
I also played around with the definitions, why not simply put the arrays into their own structs?

```c
typedef struct operator_context operator_context;
typedef struct tensor           tensor;
typedef struct array_tensor     array_input;
typedef struct array_tensor     array_output;
typedef struct array_attribute  array_attribute;
typedef int (*operator)(void *ctx);

struct tensor
{
    Onnx__TensorProto *tensor;
    operator_context  *origin;
    char              *name;
};

struct array_tensor
{
    size_t  length;
    tensor *tensor[];
};

struct array_attribute
{
    size_t                length;
    Onnx__AttributeProto *attribute[];
};

struct operator_context
{
    array_input     *input;
    array_output    *output;
    array_attribute *attribute;
    operator         operator;
};
```
this would make the specific defines more readable
```c
typedef struct array_input_TEST     array_input_TEST;
typedef struct array_output_TEST    array_output_TEST;
typedef struct array_attribute_TEST array_attribute_TEST;

struct array_input_TEST
{
    size_t  length;
    tensor *A;
    tensor *B;
};

struct array_output_TEST
{
    size_t length;
    // empty
};

struct array_attribute_TEST
{
    size_t length;
    // empty
};

typedef struct context_operator_TEST context_operator_TEST;

struct context_operator_TEST
{
    array_input_TEST     *input;
    array_output_TEST    *output;
    array_attribute_TEST *attribute;
    operator              operator;
};

void
operator_TEST(context_operator_TEST *ctx)
{
    printf("A: %s, B: %s\n", ctx->input->A->name, ctx->input->B->name);
}
```

PR #13, opened 2020-04-28
Added uncrustify code formatter

### LINKED ISSUE #12: Code formatting
Since we are adding a lot of autogenerated code and formatting this code is pure pain.
How about we add some kind of code formatter for the whole project?

- `clang-format`
  - simple config
    - common styles
    - hate it or love it approach
  - intended for c++
- `uncrustify`
  - complex config
    - almost every token transition is configurable
    - pain to configure

currently I'm evaluating both for another project and am leaning towards `uncrustify`

nopeslide:
fixes #12 

@alrevuelta 
any suggestions regarding format style is welcome

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,11 +40,26 @@ endif
 
 VARIABLE+=ONNX_CUSTOM
 HELP_ONNX_CUSTOM=use custom onnx installation
-ONNX_CUSTOM=
+ONNX_CUSTOM=third_party/onnx/onnx.build
 
 VARIABLE+=ONNX_INCLUDE
 HELP_ONNX_INCLUDE=which schemas to include
-ONNX_INCLUDE='.*'
+ifndef ONNX_INCLUDE
+ONNX_INCLUDE+="^Conv$$"
+ONNX_INCLUDE+="^Add$$"
+ONNX_INCLUDE+="^Relu$$"
+ONNX_INCLUDE+="^MaxPool$$"
+ONNX_INCLUDE+="^Reshape$$"
+ONNX_INCLUDE+="^MatMul$$"
+endif
+
+VARIABLE+=ONNX_VERSION
+HELP_ONNX_VERSION=which onnx version to use
+ONNX_VERSION=latest
+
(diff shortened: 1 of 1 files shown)


alrevuelta:
Is the `cfg` file a standard or custom one? Looks good to me. I will add the format-check to CI soon.

nopeslide:
> Is the `cfg` file a standard or custom one? Looks good to me. I will add the format-check to CI soon.

It's a custom one. the original does almost no formatting (most options are set to ignore).
Btw uncrustify moves fast. I opened a feature request issue (a missing config option) the other day and it was implemented and merged in around 2 days.
Should I add it as `third_party` submodule if we rely on it for formatting the generated code?
It would also make sense to share the uncrustify version if we have a CI check.

nopeslide:
> It would also make sense to share the uncrustify version if we have a CI check.

building it would require `g++`, `cmake` and `python` on a vanilla ubuntu

alrevuelta:
@nopeslide 
It's a custom one. the original does almost no formatting (most options are set to ignore).
Btw uncrustify moves fast. I opened a feature request issue (a missing config option) the other day and it was implemented and merged in around 2 days.
Should I add it as third_party submodule if we rely on it for formatting the generated code?
It would also make sense to share the uncrustify version if we have a CI check.

I would prefer to not add an external dependancy if it is not strictly necessary. Having the same `cfg` file, it should be backwards compatible isn't it?

nopeslide:
> I would prefer to not add an external dependancy if it is not strictly necessary. Having the same `cfg` file, it should be backwards compatible isn't it?

It throws errors on unknown configuration options. I didn't check it, but I assume it still behaves correctly. 
I asked, because integrating a format check into the CI pipeline would make it a development dependency which should be provided, shouldn't it?

alrevuelta:
> I asked, because integrating a format check into the CI pipeline would make it a development dependency which should be provided, shouldn't it?

My idea was to just use `brew install` for macOS and `apt-get` for linux and just get the latest version.

PR #29, opened 2020-06-27
generator refactoring

nopeslide:
I refactored the code generation,
made it more readable and easier to maintain.
sadly I started this before you started integrating the generation, therefore this PR needs to be rebased and identifiers need to be refactored.
I'm quite happy with code generation now (not so with my onnx wrapper, but I have no intention of touching this one soon).

What I did:
- added a dryrun option (run everything but do not touch the filesystem)
- added a Template base class for code generation
- refactored all code generation onto this new base class
- added info struct generation
- info struct can be used by non-generated check functions
- added these generic check functions
- removed all check generation


nopeslide:
@alrevuelta 
Since I need to touch almost everything I think we should look into other stuff that can be done directly. like moving the stub header into the operator.h header.

Anything else regarding the generation?

alrevuelta:
@nopeslide 
Sorry for such a late reply.

Just curious, what do you mean by "added a dryrun option (run everything but do not touch the filesystem)"?

Regarding this PR:
* You have completely freedom to modify the generators, as long as you respect the existing interfaces and so on. I'm quite happy with the integration (with some TODOs left of course) and how the architecture looks like now, so I think we can focus our efforts somewhere else. Feel like we are a bit stucked on this.
* We already talked about this, but I still don't see the need of files like `list.h` or `model.h`. I think its overcomplicated and the current solution has everything we need.
* Agree on moving the stub to `operator.h`. There is also some dead code that we can remove.

Regarding this PR and #18, are you still working on the second one? Or have you moved everything here and we can close it?

nopeslide:
> Just curious, what do you mean by "added a dryrun option (run everything but do not touch the filesystem)"?

it is a simple switch for the generator which disables file writing and directory creation. so you can generate everything and see which files would have been touched. mostly for playing around and debugging

nopeslide:
I just noticed #18 is the same dev branch I used for this PR. will close this one.

Issue #32, opened 2020-06-30
generate operator implementation templates


nopeslide:
Just thought that we could also generate the files included inside `src/operators/implementation` (i.e. `operator__onnx__relu__6__T_tensor_float.c`) from the Python script. So we can avoid manually creating them. Maybe all types is too much by now, so I would create only the `float` variants.

Quite useful also if someone wants to implement an operator. If the file and function is already in place (but empty) would be easier for someone without much knowledge of the code to implement and operator. We can abstract them from the complexity of running the Python script and so on.

_Originally posted by @alrevuelta in https://github.com/alrevuelta/cONNXr/pull/18#issuecomment-651824812_

PR #34, opened 2020-07-04
Stub generation

nopeslide:
- based on #18 
- removed weak attributes
- generates operator stub source file for each type permutation to satisfy mac&win linkers


nopeslide:
ok. mac ld seems to be happy. win ld is not. the problem with windows seems to be that the windows binary format does not support weak attributes at all.

nopeslide:
I see these options regarding windows:
1. "drop" support
   - wsl is capable of executing linux binaries
   - cygwin injects its own dll that wraps syscalls
   - wsl is much better at doing this than cygwin will ever be
2. try mingw-w64
   - the old mingw itself also had problems with weak attributes
3. try other win c compilers
   - most of them are proprietary 
4. replace weak attributes with preprocessor magic
   - maybe even Kconfig or sth similar?
   - not a fan of this one, because we would move a lot of complexity inside the build system
currently we can just omit sth while compiling and it's still a valid build (because stubs just take over)
with Kconfig we have a config dictating the the build process and if we forget to integrate it somewhere we get into trouble.

alrevuelta:
Some comments:
* Do we really need the `info` folder? Why do we need to store the attribute names? Or input/output types? Or if its variadic? Where do you plan to use all this information? Some of this information is relevant (i.e. attribute names) but IMHO it doesn't need to be accessed programatically. Also `n_attribute`, `n_input` and `n_output` are redundant. I don't think we need all these files.
* Regarding linking, weak attributes and so on. The whole point of this project is to have a runtime that is as simple as possible and that can be compiled in with every compiler. If we start to rely on compiler-dependant features that are not widely supported by all compilers, we are losing the purpose of the project. I'm not a huge fan of the following solution, but maybe we can just get rid of all the `extern`, `weak` and so on, and just have a normal function prototype. Then we generate also the implementations (files + functions) with some dummy content. Now, whenever someone wants to implement an operator they just have to replace the dummy implementation by the actual one. Continuing in #35.

nopeslide:
> Some comments:
> 
>     * Do we really need the `info` folder? Why do we need to store the attribute names? Or input/output types? Or if its variadic? Where do you plan to use all this information? Some of this information is relevant (i.e. attribute names) but IMHO it doesn't need to be accessed programatically. Also `n_attribute`, `n_input` and `n_output` are redundant. I don't think we need all these files.
> 
We do not need any of these while doing inference, they are just a byproduct of the generation and can be omitted while compiling or linking. The idea behind these informations was to provide extended debug functionality and compliance checks at runtime.
The `n_*` are not redundant since they refer to the maximum number of these objects.
I think the only redundant thing is the name.

>     * Regarding linking, weak attributes and so on. The whole point of this project is to have a runtime that is as simple as possible and that can be compiled in with every compiler. If we start to rely on compiler-dependant features that are not widely supported by all compilers, we are losing the purpose of the project. I'm not a huge fan of the following solution, but maybe we can just get rid of all the `extern`, `weak` and so on, and just have a normal function prototype. Then we generate also the implementations (files + functions) with some dummy content. Now, whenever someone wants to implement an operator they just have to replace the dummy implementation by the actual one. Continuing in #35.

Good compromise, will implement it.
I'm a little sad about the attribute stuff, but I concur.

alrevuelta:
> We do not need any of these while doing inference, they are just a byproduct of the generation and can be omitted while compiling or linking. The idea behind these informations was to provide extended debug functionality and compliance checks at runtime.

Okay, thumbs up for debugging functionalities, doesn't harm.

> The n_* are not redundant since they refer to the maximum number of these objects.

Right, I meant redundant in another way.

> Good compromise, will implement it.
I'm a little sad about the attribute stuff, but I concur.

Perfect and yep, I feel you.

nopeslide:
@alrevuelta 
implemented #32.
seems to work

alrevuelta:
@nopeslide 
Looks good to me. Just one comment. What are `operator__onnx__add__7.c`, `operator__onnx__argmax__12.c`,...? Do we expect to have something in these files or its just a bug? I mean, I expected to have `operator__onnx__add__7__TENSORTYPE.c`
In the future we can create one folder per operator so it looks less messy, but lets leave it for another patch.

nopeslide:
> @nopeslide
> Looks good to me. Just one comment. What are `operator__onnx__add__7.c`, `operator__onnx__argmax__12.c`,...? Do we expect to have something in these files or its just a bug? I mean, I expected to have `operator__onnx__add__7__TENSORTYPE.c`
> In the future we can create one folder per operator so it looks less messy, but lets leave it for another patch.

see #40

nopeslide:
@alrevuelta 
Any concerns regarding merging?

alrevuelta:
@nopeslide Go for it :D

PR #37, opened 2020-07-08
added checks if input tensor shall be skipped

nopeslide:
inputs must be skipped if they have an empty string as name


alrevuelta:
Fair check. Did you have any problems with tensors with an empty string as a name? Mobilenet perhaps? Never thought this could happen TBH.

nopeslide:
I noticed these bugs when I implemented the mobilenet operators

nopeslide:
> Never thought this could happen TBH.

we discussed this once iirc. it's part of the onnx format. if you have two optional inputs and want to specify the second, you have to set the first to an empty string

PR #38, opened 2020-07-08
Mobilenetv2 operators

nopeslide:
- based on #34  & #37 
Added an implementation of `clip_12` and `globalaveragepool_1` for the mobilenetv2 network.


alrevuelta:
Nice. I assume it is [this model](https://github.com/onnx/models/tree/master/vision/classification/mobilenet). Is it working end to end with the [test vectors](https://github.com/onnx/models/blob/master/vision/classification/mobilenet/model/mobilenetv2-7.tar.gz)? If so would be nice to add both the `onnx` and the test vectors to `test` folders and adding a test case to `include/test/models`.

nopeslide:
> Nice. I assume it is [this model](https://github.com/onnx/models/tree/master/vision/classification/mobilenet). Is it working end to end with the [test vectors](https://github.com/onnx/models/blob/master/vision/classification/mobilenet/model/mobilenetv2-7.tar.gz)? If so would be nice to add both the `onnx` and the test vectors to `test` folders and adding a test case to `include/test/models`.

Actually it's a custom one, but the onnx one shouldn't be so different.

alrevuelta:
Perfect, once we merge this I can take care of adding the onnx one.

alrevuelta:
@nopeslide Is this ready to merge?

nopeslide:
@alrevuelta yes it is, just updated the operator status

Issue #30, opened 2020-06-30
Compile as a static library to be used in other projects


alrevuelta:
Would be nice to have the project compiled as a static library so other people can easily use it in their code with just `#include connxr.h` or something like that. We could write also some nice examples in `examples` folder showing how to use it. Some work was done, but its incomplete.

codygray:
A good idea to make usage more convenient, but unfortunately, not useful.

Static libraries are toolchain-specific, so you'd have to provide one for every compiler vendor and version—and then still risk not having one for the type of esoteric embedded system where one would be using a library like this.

If you truly wanted the convenience of just `#include connxr.h`, you'd need to create a header-only library. But there's a reason these are rare, *especially* in C (as opposed to C++, just due to the semantics of the language, such as keywords like `inline`). There are many drawbacks to a header-only library. And I doubt it even fits the typical use-case of a consumer of a library like this.

alrevuelta:
@codygray Thanks for the feedback! Yep, you are right. I was thinking about an RPi or some other common hardware. It doesn't hurt to have it compiled for those. And. yeah, I wouldn't create a header-only library only for this purpose.

PR #43, opened 2020-07-14
Added elu and identity operators

mdhimes:
Hi, I'm using a modified version of cONNXr for a project, and I needed elu and identity operators.  So, I added them, and I thought you all might want those as well.  I did not include the tests that are normally performed on operators, but it's a straight-forward implementation so I'm not worried about it being wrong.  I'll add some tests if I have spare time this week.


nopeslide:
Nice, thanks for the contribution!
Testing is currently done with test data provided by the onnx project itself.
You should only need to uncomment the corresponding lines in `src/test/tests.c`.

How do you integrate cONNXr on your side? Is there something that we could do, so you can integrate cONNXr better? 
As a heads-up: there may be some breaking changes in the coming merge of #34.

alrevuelta:
Thanks! Would be great if you can share with us how you are using cONNXr. Feedback is also highly appreciated.

mdhimes:
I uncommented the relevant lines for the tests, thanks for the pointer.

We have a C code that does a lot of analytical calculations, and my research is in using NNs to reduce the compute time of those calculations.  My work to date has been using Keras in Python, so in looking for a way to deploy the model in our C code, I arrived at cONNXr.  We are still working out the exact integration of it into our main body of code, but my current approach is a function that initializes the model, and then later iteratively makes the inferences for sets of parameters.

Regarding feedback on how to make integration better -- it was already pretty easy to do, I started with the connxr.c code and altered it how I needed.  Then, I just call that function from existing code, rather than using it as an executable.  I also don't load the inputs from a .pb file, I just set up the object in the code.

Maybe something you could do to help future users would be to include an example of cONNXr integrated into another body of code.  For example maybe someone has a Pi that takes a photo when the objects in view move, so they're taking a picture every second, performing an inference, and doing something if the object moves or doesn't (maybe deletes the photo if nothing moved).  This is a kind of silly example but it would demonstrate how exactly someone can integrate cONNXr into their code.  I think it was easy enough to figure out based on the connxr.c code, but making it even easier is never a bad thing.

Also, I've experimented with multi-threading the entire codebase -- I can get some speed benefits when parallelizing the add and matmul operators, but otherwise it didn't seem to be very beneficial given the sequential nature of NNs.  The codebase has changed quite a bit since I did that test, but I'm happy to share that if you're interested.

nopeslide:
> Also, I've experimented with multi-threading the entire codebase -- I can get some speed benefits when parallelizing the add and matmul operators, but otherwise it didn't seem to be very beneficial given the sequential nature of NNs. The codebase has changed quite a bit since I did that test, but I'm happy to share that if you're interested.

I also thought about multi-threading, but like you said the benefit for a single inference is often minimal.
The largest speed-up would be parallelizing batch execution by enabling an executor to run in multiple threads for each batch element, but this needs interpretation of the tensor shape.
One of the next things I would do is rewriting the scheduler code and giving it an api in terms of `get_next_job` (to acquire the next executer) and `set_job_status` (to mark an executer as finished) so anyone can implement platform specific threading, this would at least allow for multiple threads inside a single inference. If an operator knows it can be parallelized I imagine a similar interface inside the executer.

mdhimes:
@nopeslide Yes, this is something I am experimenting with now as an alternative to multithreading the operators.  If you end up getting something like that working, please share as I'd certainly be interested.  And if I end up developing something like that, I'll be sure to let you all know.

I went ahead and resolved the merge conflicts between my fork and the latest commits.  Based on the checks it looked like some other things broke, I assume because there are now references to operators that don't exist in my branch.

alrevuelta:
@mdhimes nice insights.

> Maybe something you could do to help future users would be to include an example of cONNXr integrated into another body of code. For example maybe someone has a Pi that takes a photo when the objects in view move, so they're taking a picture every second, performing an inference, and doing something if the object moves or doesn't (maybe deletes the photo if nothing moved). This is a kind of silly example but it would demonstrate how exactly someone can integrate cONNXr into their code. I think it was easy enough to figure out based on the connxr.c code, but making it even easier is never a bad thing.

This is something that for sure we have to work on. A bit related to #30 and [this example](https://github.com/alrevuelta/cONNXr/tree/master/examples/arduino_example/Arduino_Example_ONNX) that I started but couldn't finish (its quite outdated). I totally agree with your example, would be nice to have some integration with a more real world applications and not just a single static hardcoded `.pb`.

nopeslide:
since the directory structure changed a bit, I generated the files from scratch, applied your changes and just force-pushed. this was faster than solving all merge conflicts. Since I'm taking credit now, recommit everything if you want :)

nopeslide:
rebased on #38

mdhimes:
Thanks @nopeslide I adjusted the author on that last commit.  Should be good to merge in now.

@alrevuelta I'll see if I am allowed to release a stripped-down version of what I'm working with for a real-world example.

src/operators/implementation/onnx/operator__onnx__elu__6__T_tensor_float.c
```diff
@@ -1,10 +1,58 @@
...
+
+   if (0){
+     /* TODO: Check some conditions. For example if a specific
+      * functionality is not supported */
+     return -1;
+   }
+
+   debug_print_dims(X->n_dims, X->dims);
+
+   /* TODO Alpha is not always float */
```
nopeslide, line 27: alpha is always float. see [Elu](https://github.com/onnx/onnx/blob/master/docs/Operators.md#Elu)

src/operators/implementation/onnx/operator__onnx__elu__6__T_tensor_float.c
```diff
@@ -1,10 +1,58 @@
...
+   /* Default value */
+   float alpha = 1.0;
+
+   /* If there is an attribute, its alpha */
+   if (ctx->onnx_node->n_attribute == 1){
+     alpha = ctx->onnx_node->attribute[0]->f;
+   }
```
nopeslide, line 34: could you use the `searchAttributeNyName` helper here? afaik specifying unneeded attributes is valid

src/operators/implementation/onnx/operator__onnx__elu__6__T_tensor_float.c
```diff
@@ -0,0 +1,57 @@
...
+
+   debug_print_dims(X->n_dims, X->dims);
+
+   /* Default value */
+   float alpha = 1.0;
+
+   /* If there is an attribute, its alpha */
+   if (ctx->onnx_node->n_attribute == 1){
+     alpha = ctx->onnx_node->attribute[0]->f;
+   }
```
nopeslide, line 33: I don't quite get what you mean.
If you take a look at the `Onnx__AttributeProto` declaration, you see each possible type has its own field.
This is not really covered since these declaration are generated by protobuf, the source is the only thing you have.
```suggestion
   Onnx__AttributeProto *a_alpha = searchAttributeNyName(ctx->onnx_node->n_attribute,ctx->onnx_node->attribute, "alpha");
   if (a_alpha) {
     alpha = a_alpha->f;
   }
```
Yeah the documentation is thin on this one :D
I'm also not in love with this helper, but using it, allows for quicker refactoring later on.

mdhimes: > I don't quite get what you mean.

See, for example, the conv implementation:
https://github.com/alrevuelta/cONNXr/blob/master/src/operators/implementation/onnx/operator__onnx__conv__11__T_tensor_float.c#L37
`kernel_shape` has an array called `ints` that holds the values.  I wasn't sure if this was the general approach to any attribute.  I see now that it is not.

I'll make a related adjustment to leakyrelu

mdhimes:
@nopeslide I think everything is good to go now.  Leakyrelu was also updated to use `searchAttributeNyName`.

nopeslide:
> @nopeslide I think everything is good to go now. Leakyrelu was also updated to use `searchAttributeNyName`.

Got it now^^
Thanks for pointing this out and correcting it!
We need to refactor a lot of stuff when we finished fleshing out the base architecture.

alrevuelta:
Thanks for the contribution. Quick comment to clarify some things. The following two codes are of course equivalent and in some cases they can be used indistinctly.

Code in "Way 1" can be used in the following cases:
* If there is only 1 attribute, and that attribute is mandatory, "Way 1" can be used. Not the case of `LeakyRelu` because that attribute could be empty (and the default value will be taken)
* If there are more than 1 attribute but all of them are mandatory, we can also use "Way 1". If some attributes are mandatory this can't be done.
```c
// Way 1
alpha = ctx->onnx_node->attribute[0]->f;
```

And this other way can be used in the rest of the cases.
```c
// Way 2
Onnx__AttributeProto *a_alpha = searchAttributeNyName(ctx->onnx_node->n_attribute,ctx->onnx_node->attribute, "alpha");
if (a_alpha) {
     alpha = a_alpha->f;
}
``` 

So use one or another depending on the operator. Same applies to `searchInputByName()`.

Agree that we need some refactoring. Hope to get it done once we have the architecture.

mdhimes:
Thanks for the clarification @alrevuelta !  If this approach will be used for some time, it might be good to add these details into the documentation here https://github.com/alrevuelta/cONNXr/blob/master/doc/03_Contributing.md#add-new-operator

alrevuelta:
@mdhimes You are right! Just added it to 03_Contributing.

nopeslide:
@alrevuelta 
Shouldn't we always search for an attribute? this makes the flow a lot easier, since we have a uniform handling of the attributes?

alrevuelta:
@nopeslide I guess in most of the cases we would need to search for the attributes, but in some specific cases where there is only 1 mandatory I don't see why we can't just index it. Why do you think it makes the flow easier?

Feel free to update the doc with your thoughts. I just wanted to list the possibilities that are available.

nopeslide:
> @nopeslide I guess in most of the cases we would need to search for the attributes, but in some specific cases where there is only 1 mandatory I don't see why we can't just index it. Why do you think it makes the flow easier?
> 
> Feel free to update the doc with your thoughts. I just wanted to list the possibilities that are available.
@alrevuelta 
Because we do not search/access the onnx datastructures directly but get them from our helpers, so if we decide to change names or move stuff, we only need to edit our helper.

alrevuelta:
> Because we do not have to search/access the onnx datastructures directly but get them from our helpers, so if we decide to change names or move stuff, we only need to edit our helper.
or to say it in other terms: "if you specified an interface, use it"

Okay, fair comment. We can create a more complete interface to access input/outputs/attributes if you want. Regarding "so if we decide to change names or move stuff" I don't think this will be the case with onnx data structures, but perhaps you are referring to our wrapped ones.

Lets move this discussion to an issue.

PR #48, opened 2020-07-16
rewrote transpose operator

nopeslide:
rewrote the transpose operator to handle tensors from the 5th dimension and more :D


alrevuelta:
Nice! Wasn't really proud of my spaghetti implementation. Sorry, just broke master. Fixing it.

PR #49, opened 2020-07-23
refactored/enhanced tracing

### LINKED ISSUE #31: Add traces with different levels
Currently there are 3 macros to trace information `TRACE_LEVELXX` (see `trace.h`). However they are not really used and there is a random mix between normal prints and this macros.

Task:
* Use the different `TRACE_LEVELXX` macros according to the relevancy of what is being traces. Level 0 for important information, level 1 for more detailed stuff and level 2 for very detailed traces.
* Replace the existing prints.

nopeslide:
fixes #31

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 
 VARIABLE+=TRACE_LEVEL
 HELP_TRACE_LEVEL=trace verbosity
-TRACE_LEVEL?=0
+# TRACE_LEVEL?=0
 
 VARIABLE+=BUILDDIR
 HELP_BUILDDIR=build directory
@@ -88,9 +88,11 @@ REPEAT_mobilenetv2=1
 CC=gcc
 CFLAGS+=-std=c99
 CFLAGS+=-Wall
-CFLAGS+=-g
+CFLAGS+=-g3 -gdwarf -O2
 # CFLAGS+=-Werror # CI jobs run with flag enabled
-CPPFLAGS+=-D TRACE_LEVEL=$(TRACE_LEVEL)
+ifdef TRACE_LEVEL
+CPPFLAGS+=-D "TRACE_LEVEL=$(TRACE_LEVEL)"
+endif
 
 LDFLAGS+=-g
 LDLIBS+=-lcunit
@@ -167,7 +169,7 @@ HELP_benchmark_$(1)=run $(1) benchmark
 TARGET_benchmark+=benchmark_$(1)
 benchmark_$(1): $(BENCHMARKDIR)/$(1).txt
 #Dont trace to run faster
-TRACE_LEVEL=-1
+# TRACE_LEVEL=-1
 $(BENCHMARKDIR)/$(1).txt: runtest
 	rm -f $(BENCHMARKDIR)/$(1).txt
 	mkdir -p $(BENCHMARKDIR)
--- a/doc/01_Documentation.md
+++ b/doc/01_Documentation.md
@@ -203,3 +203,50 @@ Onnx__AttributeProto *auto_pad = searchAttributeNyName(
 ```
 
 Its also important to note that each operator is usually defined for more than one data type like float, double or uint8_t. In order to address this, we have decided to define one function per data type, trying of course to share as many code as possible between different data types implementations.
+
+## Tracing
+We utilize macros to enable fine tracing of components.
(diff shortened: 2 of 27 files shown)


alrevuelta:
Looks good, few comments:
* I would "merge" all "normal" and "if" traces. So for example `TRACE_TENSOR` and `TRACE_TENSOR_IF` becomes just one named `TRACE_TENSOR` that is the conditional one. So if we want to always trace if we call it with `TRACE_TENSOR(1)`. Same applies to the other ones, so instead of having 22 tracing functions, we just have 11.
* I understand that `TRACE(1)`, `TRACE(2)` and `TRACE(3)` is equivalent to `TRACE_INFO()`, `TRACE_DETAIL()` and `TRACE_INTERNAL()`? If so, why having `TRACE()` at all? or why having all other three? I saw you changed `TRACE_DEBUG` to `INTERNAL`. TBH I still prefer having `TRACE_1`, `TRACE_2` and `TRACE_3`. The use of info, detail and internal is quite arbitrary to me so thats why I still prefer numbers.
* I'm trying to simplify the tracing because I want people to use them correctly. The more variants and complexity tracing has, the more time people will "waste" trying to understand it. Why complicating it if we can keep it simple.
* Could you add some documentation for it? Both in 01_Documentation.md and 02_Testing.md are ok to me.

nopeslide:
> Looks good, few comments:
> 
>     * I would "merge" all "normal" and "if" traces. So for example `TRACE_TENSOR` and `TRACE_TENSOR_IF` becomes just one named `TRACE_TENSOR` that is the conditional one. So if we want to always trace if we call it with `TRACE_TENSOR(1)`. Same applies to the other ones, so instead of having 22 tracing functions, we just have 11.
> 
>     * I understand that `TRACE(1)`, `TRACE(2)` and `TRACE(3)` is equivalent to `TRACE_INFO()`, `TRACE_DETAIL()` and `TRACE_INTERNAL()`? If so, why having `TRACE()` at all? or why having all other three? I saw you changed `TRACE_DEBUG` to `INTERNAL`. TBH I still prefer having `TRACE_1`, `TRACE_2` and `TRACE_3`. The use of info, detail and internal is quite arbitrary to me so thats why I still prefer numbers.
> 
was mostly testing around, and yeah it doesn't quite achieve what I wanted.
what about I simplify the interface to
```c
TRACE( level, condition, format, var... )
```

>     * Could you add some documentation for it? Either in 01_Documentation.md and 02_Testing.md are ok to me.

will do when the interface doesn't change anymore :D

another thing, I would replace the `TRACE_FILE` and `TRACE_FUNCTION` defines by env variables, so they can be set without recompiling.
so you compile with `TRACE_LEVEL=0` once as default and then can overwrite the actual tracing granularity on each execution.

```bash
make all TRACE_LEVEL=0 #activate tracing, without tracing anything`
CONNXR_TRACE_FILE=inference.c:2 ./build/connxr #run binary with finer trace settings
```

alrevuelta:
> TRACE( level, condition, format, var... )

Looks good. I just wanted to have `TRACE_XX` to enforce having only 3 levels so we don't end up having `TRACE(10)` but I like its simplicity.

> will do when the interface doesn't change anymore :D

another thing, I would replace the TRACE_FILE and TRACE_FUNCTION defines by env variables, so they can be set without recompiling.
so you compile with TRACE_LEVEL=0 once as default and then can overwrite the actual tracing granularity on each execution.

Perfect.

nopeslide:
@alrevuelta right direction?

nopeslide:
@alrevuelta If things look alright here I could merge the other PRs.
One thought I had was to maybe replace the type specific traces (`TRACE_TENSOR` and so on) by inline functions so the normal c typechecker can act upon these, but this can also be an issue for later (function type errors are more readable than macro errors for most people :D).

nopeslide:
@alrevuelta 
A btw, I did not remove the original `trace.{c,h}` files and left the inference and testing files as they are.
I only updated the operators.
should I also refactor the rest?

alrevuelta:
> A btw, I did not remove the original trace.{c,h} files and left the inference and testing files as they are.
I only updated the operators.
should I also refactor the rest?

Feel free to merge it. Inference and testing files will change quite a lot so no need to worry now.

PR #50, opened 2020-07-23
Add onnx as a python dependancy

alrevuelta:
Add onnx as a python dependancy using the "requirements.txt" and include "pip install -r requirements.txt" in the Makefile. This ensures that everytime the Python generator script is called it is ran with the version that is specified in the requirements.txt file. No need to have external dependancies in the third_party folder and no need to compile external code. Less stuff to maintain.

The --onnx option is removed from the Python script and the third_party dependancy is also removed. From now on the dependancy will be in the requirement file, currently set to 1.7.0.


alrevuelta:
@nopeslide

nopeslide:
I would leave the option to supply a custom onnx version without the need to build a python package, install it via virtualenv etc.
the script already uses any installed onnx python package if no customization is specified.

nopeslide:
one of the selling points (which got me onboard) of connxr is the easy way of integration & customization. if you want to run a custom onnx extension, connxr should be your friend.

alrevuelta:
> I would leave the option to supply a custom onnx version without the need to build a python package, install it via virtualenv etc.
the script already uses any installed onnx python package if no customization is specified.

But with this solution you can still supply a custom onnx version, just modify the version inside `requirements.txt`. Way more easy than having to change the `third_party` content or even compiling it. Of course with this we can only point to specific releases and not to specific commits, but I don't think thats needed.

My main concern with this is to enforce that when someone generates an operator, we have the same version. Lets say you have in your computer installed onnx=1.0.0 and you generate with the Makefile a new operator, but the latest onnx version is 1.6.0 and its the one that we are supporting. In that case you will be missing some operators and versions. I think we need to control that every-time someone runs `make onnx_generator` they have the version that we want.

My idea here is that we evolve in parallel with onnx, trying to support what is the latest. Then, if someone wants just v.1.0.0 or whatever, they take a subset of what we have.

> one of the selling points (which got me onboard) of connxr is the easy way of integration & customization. if you want to run a custom onnx extension, connxr should be your friend.

Agree on this :) but IMHO this commit is respecting that.

nopeslide:
> But with this solution you can still supply a custom onnx version, just modify the version inside `requirements.txt`. Way more easy than having to change the `third_party` content or even compiling it. Of course with this we can only point to specific releases and not to specific commits, but I don't think thats needed.
> 
but to for this to work you have to have onnx installed as pip package,
therefore you need to build the pip package,
therefore you need to put the pip package somewhere where it doesn't interfere with the original onnx pip package,
therefore you need to put the whole connxr repo inside a virtual env
or am I missing sth?
I see these as unnecessary steps.

> My main concern with this is to enforce that when someone generates an operator, we have the same version. Lets say you have in your computer installed onnx=1.0.0 and you generate with the Makefile a new operator, but the latest onnx version is 1.6.0 and its the one that we are supporting. In that case you will be missing some operators and versions. I think we need to control that every-time someone runs `make onnx_generator` they have the version that we want.
> 
I'm completely on your side, regarding using the pip package as default, but this is already the case.
I have nothing against removing the onnx git submodule, because you are right, it's more practical to just install it via pip.
Imho, the current behaviour should be when `CUSTOM_ONNX` is not specified, do use the installed pip package.
At the moment, this variable is set, but not used so it always defaults to pip (I'm to blame for this)

nopeslide:
Also I dislike the idea of a Makefile to manipulate system state, wouldn't a check suffice to see if the pip package is installed and a proposal how to install it, if not?

alrevuelta:
> but to for this to work you have to have onnx installed as pip package,
therefore you need to build the pip package,
therefore you need to put the pip package somewhere where it doesn't interfere with the original onnx pip package,
therefore you need to put the whole connxr repo inside a virtual env
or am I missing sth?
I see these as unnecessary steps.

But all these steps are kind of automatically ran with `pip install onnx==xx` that is run under the hood when using the `requirements.txt` file. I don't know what you mean by "interfere" with the original onnx pip package. Which one is the original one for you? For me, the original onnx package is the one that we say it is.
Regarding the venv, this is installing the specified onnx version on a system level. Agree that this is not the best idea.

> Also I dislike the idea of a Makefile to manipulate system state, wouldn't a check suffice to see if the pip package is installed and a proposal how to install it, if it is not installed?

Right, I also dislike it a bit. But if you have this check and the user doesn't have it installed, he/she will have to run exactly the same command but manually.

> Imho, the current behaviour should be when CUSTOM_ONNX is not specified, do use the installed pip package.
I think that CUSTOM_ONNX should be always specified, because we have to be sure that the scripts are being run with the version that we want.

Super random idea. Can we have or does it even make sense to have our project distributed as a Python package?

nopeslide:
> But all these steps are kind of automatically ran with `pip install onnx==xx` that is run under the hood when using the `requirements.txt` file. I don't know what you mean by "interfere" with the original onnx pip package. Which one is the original one for you? For me, the original onnx package is the one that we say it is.
>
I think we are talking about different things. I meant if you want to work with a custom onnx variant, which has no official release (i.e. bleeding edge, like I did at the beginning for the ML operators), you have to build your own pip package.  

> Regarding the venv, this is installing the specified onnx version on a system level. Agree that this is not the best idea.
> 
> > Also I dislike the idea of a Makefile to manipulate system state, wouldn't a check suffice to see if the pip package is installed and a proposal how to install it, if it is not installed?
> 
> Right, I also dislike it a bit. But if you have this check and the user doesn't have it installed, he/she will have to run exactly the same command but manually.
> 
Either we let the user decide how to install onnx, or we create our own venv and install onnx there. 
 
> Super random idea. Can we have or does it even make sense to have our project distributed as a Python package?
>
Without python bindings? :D I think other projects like microsofts onnxruntime are quite pythonic enough.
and since we don't target python (afaik, tell me if I got sth wrong :D), can't see the benefit here.

alrevuelta:
@nopeslide Coming back to this, can we agree then on removing the external dependancy (as I did) and don't modify the onnx version within the Makefile (as you suggested) ? So we just check which version the user has, and if its not the one that we want, we don't run the script?

nopeslide:
@alrevuelta 
> Coming back to this, can we agree then on removing the external dependancy (as I did)

yes

> and don't modify the onnx version within the Makefile (as you suggested)?

the ONNX version inside the Makefile is as far as I remember not the release version number but the IR version number. 
so the value "latest" should always grab the latest operator versions.
With the new coming build system I plan to split the `onnx_generator` into one tool that only generates operator templates (sth like `operator_template ai.onnx Add 12` and you get all files regarding this operator version) and a second which generates all sets known to onnx (which is only used to update the sets with the latest onnx version)

> So we just check which version the user has, and if its not the one that we want, we don't run the script?

I would prefer a virtual env at this point, since we will also add more python stuff regarding testing etc.
In this venv we simply install everything we need with pip

nopeslide:
@alrevuelta 
How about we create a venv in `/build/venv` and manage it with make?
We would need to create it once manually and then use `pip freeze` to export all necessary things and save these in `/requirements.txt`. make just recreates it if it doesn't exist and make uses the venv when starting our scripts?
we would need to introduce a second clean stage (template should already be capable of doing so), where `make clean` removes c related files and `make distclean` removes everything

alrevuelta:
> How about we create a venv in /build/venv or /venv and manage it with make?
We would need to create it once manually and then use pip freeze to export all necessary things and save these in /requirements.txt. make just recreates it if it doesn't exist and make uses the venv when starting our scripts?
we would need to introduce a second clean stage (template should already be capable of doing so), where make clean removes c related files and make distclean removes everything (including the venv)

Looks like a good idea. Here it is.

Makefile
```diff
@@ -246,7 +246,8 @@ format-check:
 HELP_onnx_generator=generate various onnx sources and headers
 TARGET+=onnx_generator
 onnx_generator:
-	pip install -r requirements.txt
+	python -m venv venv
+	. venv/bin/activate; pip install -r requirements.txt
```
nopeslide: would make the venv as a standalone target in preparation for other python related stuff and to not run pip each time

alrevuelta: I like having it here so we are 100% sure that the correct version is used. Using `venv` does not remove or override the existing virtual environment (if any), so I don't think its a problem. If the environment already exists it will continue and if the version is already installed, nothing will happen.

nopeslide: but we have no way of modifying the venv prior the execution.
let's say we want to test something and install another onnx version inside the venv, it will get overwritten without mercy.
I see the `requirements.txt` more as a setup configuration, not as a execution configuration.
no big issue on my side though

alrevuelta: Will close this. If in the future we need to install other versions, we can discuss it further.

PR #53, opened 2020-08-02
refactored / enhanced conv operator

nopeslide:
- based on #49 
- added generic implementation
- added group functionality
- added dilation functionality
- added string utils
  - strdup
  - strndup
  - strnlen

### DIFF
--- a/src/operators/implementation/onnx/operator__onnx__conv__11__T_tensor_float.c
+++ b/src/operators/implementation/onnx/operator__onnx__conv__11__T_tensor_float.c
@@ -2,8 +2,11 @@
 #include <stdlib.h>
 #include <string.h>
 #include <math.h>
-#include "tracing.h"
+#include <stdbool.h>
+#include "trace.h"
 #include "utils.h"
+#include "tracing.h"
+#include "index.h"
 
 operator_status operator__onnx__conv__11__T_tensor_float(
     node_context *ctx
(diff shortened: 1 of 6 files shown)


src/operators/implementation/onnx/operator__onnx__conv__11__T_tensor_float.c
```diff
@@ -2,162 +2,331 @@
...
-    //a->n_dims == b->n_dims
-    //a->dims[i] == b->dims[i]
-    // dilations is hardcoded ?
-    fprintf(stderr, "Not implemented\n");
-    exit(-1);
+  TRACE_ENTRY(1);
+
+  TRACE_NODE(2, true, ctx->onnx_node);
+
+  Onnx__TensorProto *t_X = searchInputByName(ctx, 0);
```
alrevuelta, line 19: I liked having the same names as onnx has, `X`, `W`, so you can recognise the variable on a quick glance

nopeslide: I had a naming problem with the attributes.
Since all attributes are optional, I couldn't use them directly.
I prefixed the onnx attributes with `a_` so I can refer to the actual values without the prefix.
`a_kernel` is the onnx attribute, `kernel` the usable value.
To be consistent I also added a tensor prefix `t_`.

PR #55, opened 2020-08-03
Operator dirs

### LINKED ISSUE #41: directory structure for operators
In the future we can create one folder per operator so it looks less messy, but lets leave it for another patch.

_Originally posted by @alrevuelta in https://github.com/alrevuelta/cONNXr/pull/34#issuecomment-657605954_

### LINKED ISSUE #47: rename our default onnx domain `onnx` to `ai.onnx`
I just noticed the domains onnx uses are all prefixed by `ai.onnx`. so we should rename our default domain `onnx` to `ai.onnx`

nopeslide:
- fixes #41 
- fixes #47 
- based on #53


alrevuelta:
No need to add it to this patch, but just came up with the idea of adding also one `__utils.h-c` file per folder. All operators data types version will share some code, so perhaps we can enforce it by having this file. What do you think?

nopeslide:
> No need to add it to this patch, but just came up with the idea of adding also one `__utils.h-c` file per folder. All operators data types version will share some code, so perhaps we can enforce it by having this file. What do you think?

would not enforce it. with the build system overhaul I intend to give each operator its own Makefile so it can compile everything as it sees fit.

alrevuelta:
> would not enforce it. with the build system overhaul I intend to give each operator its own Makefile so it can compile everything as it sees fit.

Roger that.

PR #66, opened 2020-08-06
Add benchmarking for mobilenetv2

alrevuelta:
Add benchmarking for mobilenetv2


nopeslide:
could you remove the `.DS_Store` files?

.gitignore
```diff
@@ -2,3 +2,4 @@ build
 profiling
 benchmarks
 dump.txt
+.DS_Store
```
nopeslide, line 5:
```suggestion
**/.DS_Store
```

alrevuelta: Looks like works without it, but ok :)

nopeslide: ok. was not sure of the default behaviour of .gitignore here

nopeslide:
could you squash the the `.DS_Store` related commits so it does not clutter our history?

PR #67, opened 2020-08-06
Crawl previous onnx tests

alrevuelta:
This PR introduces a new way of testing the operators, by relying both on onnx default test cases and also the possibility of easily writing custom test cases. So testing is divided into two:

* **Official onnx tests**: These tests are the ones that [onnx provides](https://github.com/onnx/onnx/tree/master/onnx/backend/test/data/node). Note that onnx does not keep track of opset versions, so in a given release there are only test cases for that release. If an operator changes, that test is overriden by the new one. This PR addresses that, by crawling all onnx releases and organising the testcases by domain and version. Note that there are [ongoing conversations](https://github.com/onnx/onnx/issues/2912) with onnx regarding this topic.

* **Custom tests**: On the other hand, this PR introduces the possibility of writing custom tests in Python (as onnx does) and generating a model with a set of inputs and expected outputs. See examples in node_custom_case. To generate the expected output, onnx runtime is used, so just the input needs to be provided. With this, there is no excuse, a new test can be written in seconds, and its model can be generated by running a simple script.

The tests are organised in two folders with the same structure:
* `node_official_data` that contains the **all** official onnx tests, organized by domain and version. Note that this can be generated by `get_official_tests.py` script, which makes it easy to keep in synch if a new onnx version is released.
* `node_custom_data`, that contains the **custom** tests written by us. These tests are generated using the Python files inside `node_custom_case`. The script that generates the models is `generate_custom_tests.py`.

The file structure of both folders will look like:
image


nopeslide:
is it necessary to have these operator tests release-based? can't we extract the operator version specific tests?

alrevuelta:
> is it necessary to have these operator tests release-based?

I'm not sure I understand what you mean. Onnx does not keep track of the operator versions, they just create releases that contains test cases for that given release, that can be mapped to a given opset version.

> can't we extract the operator version specific tests?
afaik thats not possible. Onnx doesn't "organise" the tests per operator version. For each release they override all previous testcases so they comply with the new operator version. So to the best of my knowledge, and as suggested here: onnx/onnx#2912 this is the only way to do it. If in the future onnx changes (which we can help) this might get simpler.

nopeslide:
@alrevuelta 
What I meant was, If we have onnx versions and a mapping from onnx version to opset and from opset to op version, couldn't we label the test of each operator with its operator version corresponding to these mappings?

alrevuelta:
@nopeslide 

You mean that instead having a folder with this structure:
image

we could have just one folder (`onnx_backend`) with all test cases like:
* test_add_3
* test_add_7
* test_add_12
* test_conv..._3
* test_vonv_...11
* ...

nopeslide:
@alrevuelta 
exactly
we would have no duplicates and save a lot of files

alrevuelta:
> exactly
we would have no duplicates and save a lot of files

Sure I can do this, but the number of files would be the same. They would just be inside the same folder and not spread across different ones . But don't you think its nice to have them organised by opset version? Lets say we want to run the tests of a given opset. If we don't organise them by folders, we would need to check every test name to see the version it belongs. I also like having it organised by folders because it maps to a specific onnx release, which makes it easy to track.

nopeslide:
> Sure I can do this, but the number of files would be the same. They would just be inside the same folder and not spread across different ones . But don't you think its nice to have them organised by opset version? Lets say we want to run the tests of a given opset. If we don't organise them by folders, we would need to check every test name to see the version it belongs. I also like having it organised by folders because it maps to a specific onnx release, which makes it easy to track.

I misunderstood you then. I thought each release has the tests for all operators, which would imply a lot of duplicates, since not all operators change versions between releases.

alrevuelta:
> I misunderstood you then. I thought each release has the tests for all operators, which would imply a lot of duplicates, since not all operators change versions between releases.

Actually you are right. Now we have lots of duplicates. I was working under a wrong assumption.

Looks like if a given operator doesn't change for the next release, its test data is still included. So test data for lets say opset 5 will include also the tests for previous operator versions that were unchanged.

This makes the test data of opset `n` a subset of `n+1`, and what we really want is just the "delta". So we need to make sure that `n+1` doesn't contains what we already have in `n`. I was thinking about hashing the folders and using the hash as an identifier to avoid duplicates. Any other approach perhaps?

nopeslide:
> This makes the test data of opset `n` a subset of `n+1`, and what we really want is just the "delta". So we need to make sure that `n+1` doesn't contains what we already have in `n`. I was thinking about hashing the folders and using the hash as an identifier to avoid duplicates. Any other approach perhaps?

just iterate over all onnx versions, starting from the lowest, and each time split all tests according to the operator version and overwrite existing ones.
afterwards you have your unique tests and each one of them has the most up-to-date version.

alrevuelta:
> just iterate over all onnx versions, starting from the lowest, and each time split all tests according to the operator version and overwrite existing ones.

I got rid of the duplicates but there is still something tricky.

For example, the operator `Abs` has only 3 versions (13, 6, 1), so I was expecting to have only test cases for the Abs operator in the releases that correspond to those ones. I was also expecting to have three test cases with three models, where the imported opset was 13, 6 and 1 for each one. But its not like that.

Due to some modifications or updates, there are models testing `Abs` that imports opset 8 or 9. So the real situation is something like:

* opset 1 contains Abs
* opset 6 contains Abs
* opset 8 contains Abs (but Abs didn't change in opset 8)
* opset 9 contains Abs (but Abs didn't change in opset 9)
* opset 13 contains Abs

So in this case, the `Abs` tests for opset 8 and 9 refer also to opset 6, because `Abs` didn't change in this opsets. So we have two solutions:
* Either we accept that Abs opset 6 tests are (6+8+9)
* Or we get rid of some of them and we take only 6 or 9 as the tests.

I would go for option 1, since it doesn't involve removing any tests and this tests change (that refer to the same opset) might add some value.

And btw, the structure looks now like this (not complete yet)
image

@nopeslide thoughts?

alrevuelta:
@nopeslide I just updated the PR text, could you have a look? Apart from crawling all onnx tests, I added the possibility of writing custom test cases, which I think is really important. I have some TODOs left, hope to finish it during the weekend.

nopeslide:
@alrevuelta is #85 replacing this PR?

alrevuelta:
> @alrevuelta is #85 replacing this PR?

Yes, will close it. I think #85 is a better way of testing the operators, as we can write custom test cases. This will ofc take more time that just reusing onnx test cases, but since onnx tests are not organised per opset version and crawling them is a hell I think its a better option. Its also a good idea to force writing custom test cases, as onnx ones are not that complete.

If in the future onnx organices the tests, we can always switch to them. I had some talks with them and there was an interest in doing this, but the changes are non trivial and unfortunately I don't have time to allocate to it. Left a draft [PR](https://github.com/onnx/onnx/pull/3020) that might be useful if someone in the future wants to take over.

PR #72, opened 2020-09-03
Implemented the Sigmoid function for floats

bjornite:


nopeslide:
Thanks for contributing!
Could you also uncomment the corresponding operator tests in `/src/test/tests.c`?

alrevuelta:
Thanks for your contribution! Couple of comments:

* Could you uncomment from `tests.c` file `test_sigmoid` and `test_sigmoid_example` tests? These tests are taken directly from ONNX backend set of tests, and by uncommenting them your operator will be tested. Note that this tests only cover float type. We are working on new ways to improve testing though, see #69.

* Note that we are preparing a patch #57 that hopefully will be merged soon. This patch changes a bit the structure of the operators (and will requiere this PR to be modified). TLDR; That PR avoids some boilerplate code and moves out of the specific operator (float, double,...) some code to a common function. If you are interested, feel free to take a look to for example [this](https://github.com/alrevuelta/cONNXr/pull/57/files#diff-9e7ccc8ac6a87cf459e76c443ecf68c0).

Since you have contributed with an operator, you might have some specific feedback about how the process was, do you? We are trying to make contributing as easy as possible, having a modular and solid architecture where a new operator can be easily implemented without knowing much about what is under the hood, so any feedback is appreciated.

@nopeslide Any comments?

Thanks!

bjornite:
Made the changes to enable the tests now. Thanks for the pointers.

@alrevuelta I found it straight forward to implement the operator. You've done a great job making it easy to contribute in my opinion :)

alrevuelta:
@bjornite Thanks for the contribution!

PR #73, opened 2020-09-11
Build as shared + Migrate testing to Python

### LINKED ISSUE #60: Build connxr as a shared library, to enable python bindings for testing
- Python should be our way to test things
- Having python bindings would simplify testing a lot
  - can use the onnx python library, numpy, onnxruntime etc
  - no need to generate files, everything can be done in memory
- Central script that does all the testing
  - can generate the operator status overview (#44) and more

alrevuelta:
Fix #60 

* Migrates all testing of operators and models to Python, by building a shared library and interfacing to it from Python.
* Migrates the benchmarking of models to Python. Now there is no need to use files.
* Simplify the Makefile.
* Proof of concept before using new test data (organised by opset version and domain in #67)


nopeslide:
Nice start
> static library

shouldn't this be dynamic/shared in the PR text?

tests/test_models.py
```diff
@@ -0,0 +1,53 @@
...
+        self.assertTrue(test_model(b"super_resolution",
+                                   b"test/super_resolution/super_resolution.onnx",
+                                   b"test/super_resolution/test_data_set_0",
+                                   1, 1) > 0);
+
+    def test_tinyyolov2(self):
+        self.assertTrue(test_model(b"tinyyolov2",
+                                   b"test/tiny_yolov2/Model.onnx",
+                                   b"test/tiny_yolov2/test_data_set_0",
+                                   1, 1) > 0);
```
nopeslide, line 48: how about we create a skeleton class for these things?
something like:
```python
class TestModel():
  connxr = CDLL('build/connxr.so')
  def test_model(self):
    result = test_model_function(self.id, self.path, self.io_path, self.n_inputs, self.n_outputs)
    self.assertTrue(result)
    print("Inference time", result)

class TestMnist(TestModel, unittest.TestCase):
  id = b"mnist"
  path = b"test/mnist/model.onnx",
  io = b"test/mnist/test_data_set_0",
  n_inputs = 1
  n_outputs = 1
```

alrevuelta: Sure

alrevuelta:
> shouldn't this be dynamic/shared in the PR text?

You are absolutely right.

I'm gonna need a bit of help with the Makefile, might be doing really nasty stuff. Can you have a look? The main problem now is that we need to compile with the `-fpic` flag all the code, but we only want this for the shared library. Should be have a target like `compile_shared` or something like that?

nopeslide (approved):

PR #83, opened 2021-01-11
Add compatibility for older compilers

### LINKED ISSUE #81: Tracing not compatible with C99
Tracing is not compatible with C99.
On the other hand, as already discussed in some PR with @nopeslide I think we should also simplify tracing. We have lots of functions and macros, making it complex to use and enforce its correct use.

Coderitter-GmbH:
Closes #81

Older versions of gcc don't handle `__VA_OPT__` correctly. This pull request works around that, with removing `__VA_OPT__`.


alrevuelta:
Looks good to me. @nopeslide ?

PR #85, opened 2021-02-14
Add custom testcases generator

alrevuelta:
This PR adds the possibility of adding custom test cases for operators in a similar manner as its done in [onnx](https://github.com/onnx/onnx/tree/master/onnx/backend/test/case/node). The idea is to generate all `model.onnx` and set of test vectors (input + output) from Python code, as its done in onnx.

For example, the following code will generate a set of inputs/outputs and a model.onnx that will test the `Add` operator on its version 7. Note that multiple `export` methods can be used to define different tests for the same operator. This schema is taken from onnx repo, with the only difference that `opset` is also provided, and the output test vector is calculated by running inference with the official onnx runtime.

```python
class Add(Base):

    @staticmethod
    def export():  # type: () -> None
        node = onnx.helper.make_node(
            'Add',
            inputs=['x', 'y'],
            outputs=['sum'],
        )

        x = np.random.randn(3, 4, 5).astype(np.float32)
        y = np.random.randn(3, 4, 5).astype(np.float32)
        expect(node,
               inputs=[x, y],
               name='test_add_custom',
               opset_import=7,
               domain="ai.onnx")

```

Some notes:
* The main rationale behind this PR is that in spite of onnx provides test vectors for operators, they are not defined per operator version, and they don't keep track of it. This means that if we just take the latest onnx official test vectors, we would only test the latest opset version, since older ones are overiden.
* PR #67 aimed to address the previous bullet by crawling previous test vectors, but the solution is not very elegant and tons of tests are repeated.
* Almost all the code that generates the models is taken from onnx one. See [this](https://github.com/onnx/onnx/blob/master/onnx/backend/test/case/node/__init__.py) for reference.

So whenever a new operator is implemented, to make sure its working as expected we will demand also a Python code that tests that new operator for a given opset and type (see `Add` example). The tests are organised per domain and opset (see folders).

Once the Python code is places, running the following command will place in `test_data/node_custom_data` the `model.onnx` and a set of input/output test vectors that can be used to for testing (PR to come).

```console
make generate_custom_tests
```


alrevuelta:
@nopeslide Could you have a look? I recycled most of this from #67 . Most of the code is reused from onnx, so `generate_custom_tests.py` and `__init__.py` have very minor changes. What I'm interested in is in the idea of having our own way of easily adding testcase per operator and opset version, so there are no excuses for adding new tests.

nopeslide:
@alrevuelta
It looks quite nice.
I don't get everything that happens in this (quite large) commit,
but the node implementations look easy to implemented.

PR #88, opened 2021-03-08
Added cunit install note to README

nopeslide:


alrevuelta (approved):
Perfect. As a side note, we are not really using now CUnit, as all the testing was moved to Python, but I left it in place for the future. So the plan from now on is to use CUnit to test helper functions and stuff like that, leaving models/operators to Python.

nopeslide:
> Perfect. As a side note, we are not really using now CUnit, as all the testing was moved to Python, but I left it in place for the future. So the plan from now on is to use CUnit to test helper functions and stuff like that, leaving models/operators to Python.

doesn't it make sense to move these function tests also to python?

alrevuelta:
> doesn't it make sense to move these function tests also to python?

I mean, I think we would need CUnit to run some tests on C side, thats why I left it. Not sure yet what would be though. Do you agree that we might need CUnit in the future? I don't see all tests running on Python side, but maybe I'm wrong.

nopeslide:
I just don't want to manage multiple test suites^^

alrevuelta:
> I just don't want to manage multiple test suites^^

Yep, thumbs up for that. FWIW, when I migrated CUnit to Python everything was quite smooth when dealing with simple datatypes (strings, ints), but things can get very complicated if the input to the function that you are testing is an array, or a type that is defined in C. I would like to manage just one test suit, but experience with Python tells me that things can get tricky.

nopeslide:
even if you use `ctypes` and `struct`?

PR #91, opened 2021-03-09
ConvTranspose2d Operator

### LINKED ISSUE #79: Add support for ConvTranspose2d
To support convolutional autoencoder models, this library should support ConvTranspose2d and pass the ONNX tests.

Coderitter-GmbH:
First Implementation of the ConvTranspose2d Operator. Its not support every combination of parameters but it passes the most tests.

fixes #79


src/operators/ai.onnx/ConvTranspose/11/execute_operator__ai_onnx__convtranspose__11__T_tensor_float.c
```diff
@@ -0,0 +1,193 @@
...
+
+    /* UNCOMMENT AS NEEDED */
+
+    Onnx__TensorProto *i_X = searchInputByName(ctx, 0);
+    Onnx__TensorProto *i_W = searchInputByName(ctx, 1);
+    Onnx__TensorProto *i_B = searchInputByName(ctx, 2);
+
+    // TRACE_TENSOR(2, true, i_X);
+    // TRACE_TENSOR(2, true, i_W);
+    // TRACE_TENSOR(2, B, i_B);
```
nopeslide, line 44: why not use these?

src/operators/ai.onnx/ConvTranspose/11/execute_operator__ai_onnx__convtranspose__11__T_tensor_float.c
```diff
@@ -0,0 +1,193 @@
...
+    // int64_t* strides = op_ctx->strides;
+
+    // TRACE_VAR(2, true, auto_pad, "\"%s\"");
+    // TRACE_ARRAY(2, true, dilations, , n_dilations, "%" PRId64);
+    // TRACE_VAR(2, true, group, "%" PRId64);
+    // TRACE_ARRAY(2, true, kernel_shape, , n_kernel_shape, "%" PRId64);
+    // TRACE_ARRAY(2, true, output_padding, , n_output_padding, "%" PRId64);
+    // TRACE_ARRAY(2, true, output_shape, , n_output_shape, "%" PRId64);
+    // TRACE_ARRAY(2, true, pads, , n_pads, "%" PRId64);
+    // TRACE_ARRAY(2, true, strides, , n_strides, "%" PRId64);
```
nopeslide, line 73: also unused

src/operators/ai.onnx/ConvTranspose/11/execute_operator__ai_onnx__convtranspose__11__T_tensor_float.c
```diff
@@ -0,0 +1,193 @@
...
+
+    Onnx__TensorProto *o_Y = searchOutputByName(ctx, 0);
+    printf("%p\n", o_Y->float_data);
```
nopeslide, line 76: please do not use printf, but the tracing functions

src/operators/ai.onnx/ConvTranspose/11/execute_operator__ai_onnx__convtranspose__11__T_tensor_float.c
```diff
@@ -0,0 +1,193 @@
...
+        padEndX = a_pads->ints[3];
+    }
+
+    const int outputSizeX = calcOutputSize(inputSizeX, kernelSizeX, strideX, dilationsX, 0, 0);
+    const int outputSizeY = calcOutputSize(inputSizeY, kernelSizeY, strideY, dilationsY, 0, 0);
+
+    const int outputSizePaddedX = calcOutputSize(inputSizeX, kernelSizeX, strideX, dilationsX, padStartX, padEndX);
+    const int outputSizePaddedY = calcOutputSize(inputSizeY, kernelSizeY, strideY, dilationsY, padStartY, padEndY);
+
+    //if badding is needed we need an extra tensor and applay padding later
```
nopeslide, line 129: typos

src/operators/ai.onnx/ConvTranspose/11/execute_operator__ai_onnx__convtranspose__11__T_tensor_float.c
```diff
@@ -0,0 +1,193 @@
...
+    float* output = NULL;
+    if(outputSizeX * outputSizeY != outputSizePaddedX * outputSizePaddedY) {
+        output = (float*)malloc(outputSizeX * outputSizeY * outputChannels * sizeof(float));
+    } else {
+        output = o_Y->float_data;
+    }
```
nopeslide, line 135: but why? why not have the padding in the index calculation?

src/operators/ai.onnx/ConvTranspose/11/execute_operator__ai_onnx__convtranspose__11__T_tensor_float.c
```diff
@@ -0,0 +1,193 @@
...
+        }
+    } else {
+        for(int c=0; c<outputChannels; c++) {
+            for(int y=0; y<outputSizeY; y++) {
+                for(int x=0; x<outputSizeX; x++) {
+                    output[calcArrayPos3D(x,y,c,outputSizeX, outputSizeY)] = i_B->float_data[c];
+                }
+            }
+        }
+    }
```
nopeslide, line 154: why do this here and not in the actual convolution?

nopeslide (changes requested):

nopeslide:
Thanks for the PR!
I noticed a few things and commented accordingly.

src/operators/ai.onnx/ConvTranspose/11/execute_operator__ai_onnx__convtranspose__11__T_tensor_float.c
Coderitter-GmbH: Because the convolution is a sum calculated with += in a loop and we need a starting value which is either 0 or the bias value

src/operators/ai.onnx/ConvTranspose/11/execute_operator__ai_onnx__convtranspose__11__T_tensor_float.c
Coderitter-GmbH: That would involve multiple if statements to not write over the border. The problem with the padding in ConvTranspose2d is that the padding makes the result smaller (the lines get cut off). Any ideas how we could solve this better?

src/operators/ai.onnx/ConvTranspose/11/execute_operator__ai_onnx__convtranspose__11__T_tensor_float.c
nopeslide: thanks, I forgot you loop for each channel over the same output cell m(
would still modify this to sth like
```c
        for(int c=0; c<outputChannels; c++) {
            float bias = i_B?i_B->float_data[c]:0;
            for(int y=0; y<outputSizeY; y++) {
                for(int x=0; x<outputSizeX; x++) {
                    output[calcArrayPos3D(x,y,c,outputSizeX, outputSizeY)] = bias;
                }
            }
        }
    }

```

Coderitter-GmbH: I didn't want if statements inside the loop but yes its only one if per channel so I can change it

src/operators/ai.onnx/ConvTranspose/11/execute_operator__ai_onnx__convtranspose__11__T_tensor_float.c
nopeslide: > That would involve multiple if statements to not write over the border.

Do you think this will be a performance problem? I think the kernel detection in gcc should optimize this pretty well.
At least I would allocate this buffer inside the prepare function, so the executer does no malloc.

src/operators/ai.onnx/ConvTranspose/11/execute_operator__ai_onnx__convtranspose__11__T_tensor_float.c
```diff
@@ -0,0 +1,193 @@
...
+    return outputChannel * height * width + y * width + x;
+}
+
+//transformes a 4d pos into a 1d flat array pos
+int calcArrayPos4D(int x, int y, int outputChannel, int intputChannel, int width, int height, int nOfOutputChannels) {
+    return intputChannel * nOfOutputChannels *height * width 
+           + outputChannel * height * width 
+           + y * width
+           + x;
+}
```
nopeslide, line 25: would inline these

src/operators/ai.onnx/ConvTranspose/11/execute_operator__ai_onnx__convtranspose__11__T_tensor_float.c
nopeslide: why no if statements inside the loop? sounds like you are trying to solve performance issues (did you have any?)

src/operators/ai.onnx/ConvTranspose/11/execute_operator__ai_onnx__convtranspose__11__T_tensor_float.c
Coderitter-GmbH: As far I know is Inline is only a suggestion, if the function is small enough the compiler will inline it even without the inline statement

nopeslide: `static inline` is your friend here.
gcc may inline these on its own, but by saying `static inline` you tell gcc to inline them and not generate a symbol, therefore it has to inline these and produces no callable assembler section

src/operators/ai.onnx/ConvTranspose/11/execute_operator__ai_onnx__convtranspose__11__T_tensor_float.c
Coderitter-GmbH: I don't know anything about the kernel detection in gcc. The final convolution are six for loops for a 4d tensor adding additional statements have the potential for a huge impact to make it slower. Maybe we should benchmark it.

src/operators/ai.onnx/ConvTranspose/11/execute_operator__ai_onnx__convtranspose__11__T_tensor_float.c
Coderitter-GmbH: I didn't had performance issues but neither did I want to make the program unnecessary slower. In this case here it doesn't matter.

src/operators/ai.onnx/ConvTranspose/11/execute_operator__ai_onnx__convtranspose__11__T_tensor_float.c
nopeslide: > You could use `__attribute__((__always_inline__))` if you want force it

this is bad, since it also inlines if `-O0` is set

> only visible in this scope

not really, `static` refers to internal linkage in this translation_unit (file) and that means no symbol, since symbols are needed for external linkage.
please read the c99 standard and gcc manual on this topic if you don't want believe me :)
regarding the combination from current gcc [manual](https://gcc.gnu.org/onlinedocs/gcc/Inline.html)
```
When a function is both inline and static,
if all calls to the function are integrated into the caller,
and the function’s address is never used,
then the function’s own assembler code is never referenced. 
[...]
When an inline function is not static,
then the compiler must assume that there may be calls from other source files;
since a global symbol can be defined only once in any program,
the function must not be defined in the other source files,
so the calls therein cannot be integrated.
Therefore, a non-static inline function is always compiled on its own in the usual fashion. 
```

src/operators/ai.onnx/ConvTranspose/11/execute_operator__ai_onnx__convtranspose__11__T_tensor_float.c
nopeslide: yeah I was more concerned about copying these large amounts of data as about the padding edge case.
but whatever, if it is slow we can optimize it later on

nopeslide: Isn't this operator very familiar with the conv operator? we could benchmark your implementation against our already existing one
[execute_operator__ai_onnx__conv__11__T_tensor_float](https://github.com/alrevuelta/cONNXr/blob/master/src/operators/ai.onnx/Conv/11/execute_operator__ai_onnx__conv__11__T_tensor_float.c)

Coderitter-GmbH: It is kinda similar because it does the exact opposite than conv

src/operators/ai.onnx/ConvTranspose/11/execute_operator__ai_onnx__convtranspose__11__T_tensor_float.c
Coderitter-GmbH: Should I prefer op_ctx over searchAttributeNyName?

nopeslide: the best way would be to resolve attributes once with `searchAttributeNyName` and save them to `op_ctx` (or replace them with a default) so the executer doesn't have to resolve these again.

Coderitter-GmbH: Now i get it. I will adopt my implementation

Coderitter-GmbH:
I still have to:

- remove the unused function 
- use op_ctx
- use trace

nopeslide:
> To use default_dilations, default_pads and default_strides I have to allocate memory. Doesn't that cause memory leak, because later I don't know if op_ctx->pads contains the pointer from a_pads->ints or my self allocated memory (ARRAYDUP)

you have multiple option to solve this:
- always dup
- mark dup as dup with extra flag or set to NULL to use onnx one
- compare pointers when freeing

Coderitter-GmbH:
> I think I would some day prefer the comparison, since it's simple and moves the annoying stuff into the free function
> but atm I would simply dup everything. we can always start from there for optimizations, especially since the free function is not utilized at this point

Okay dup everything it is for now

Coderitter-GmbH:
Can you look over it again?

src/operators/ai.onnx/ConvTranspose/11/execute_operator__ai_onnx__convtranspose__11__T_tensor_float.c
```diff
@@ -0,0 +1,149 @@
...
+    const int dilationsX = dilations[1];
+    const int dilationsY = dilations[0];
+
+    const int padStartY = pads[0];
+    const int padStartX = pads[1];
+    const int padEndY = pads[2];
+    const int padEndX = pads[3];
+
+    const int outputSizeX = calcOutputSize(inputSizeX, kernelSizeX, strideX, dilationsX, padStartX, padEndX);
+    const int outputSizeY = calcOutputSize(inputSizeY, kernelSizeY, strideY, dilationsY, padStartY, padEndY);
```
nopeslide, line 98: isn't this equal to the output tensors dimension?
couldn't you just use `o_Y->dims`?

src/operators/ai.onnx/ConvTranspose/11/execute_operator__ai_onnx__convtranspose__11__T_tensor_float.c
```diff
@@ -0,0 +1,149 @@
...
+        for(int c=0; c<outputChannels; c++) {
+            for(int inputPosY=0; inputPosY<inputSizeY; inputPosY++) {
+                for(int inputPosX=0; inputPosX<inputSizeX; inputPosX++) {
+                    float _input = input[calcArrayPos3D(inputPosX, inputPosY, i, inputSizeX, inputSizeY)];
+
+                    for(int kernelPosX=0; kernelPosX<kernelSizeX; kernelPosX++) {
+                        int x = inputPosX*strideX+kernelPosX*dilationsX - padStartX;
+                        if(x < 0) {
+                            continue;
+                        } else if (x >= outputSizeX) {
```
nopeslide, line 123: make these to one condition,
it's easier to read

src/operators/ai.onnx/ConvTranspose/11/execute_operator__ai_onnx__convtranspose__11__T_tensor_float.c
```diff
@@ -0,0 +1,149 @@
...
+                            continue;
+                        }
+
+                        for(int kernelPosY=0; kernelPosY<kernelSizeY; kernelPosY++) {
+                            int y = inputPosY*strideY+kernelPosY*dilationsY - padStartY;
+
+                            if(y < 0) {
+                                continue;
+                            } else if (y >= outputSizeY) {
```
nopeslide, line 132: s.a.

src/operators/ai.onnx/ConvTranspose/11/prepare_operator__ai_onnx__convtranspose__11.c
```diff
@@ -0,0 +1,182 @@
...
+    const int padStartY = op_ctx->pads[0];
+    const int padStartX = op_ctx->pads[1];
+    const int padEndY = op_ctx->pads[2];
+    const int padEndX = op_ctx->pads[3];
+
+    int outputSizeX = calcOutputSize(inputSizeX, kernelSizeX, strideX, dilationsX, padStartX, padEndX);
+    int outputSizeY = calcOutputSize(inputSizeY, kernelSizeY, strideY, dilationsY, padStartY, padEndY);
+
+    o_Y->n_dims       = i_X->n_dims;
+    o_Y->dims         = malloc(o_Y->n_dims * sizeof(int64_t));
```
nopeslide, line 154: `TRACE_FATAL` for this malloc

src/operators/ai.onnx/ConvTranspose/11/prepare_operator__ai_onnx__convtranspose__11.c
```diff
@@ -0,0 +1,182 @@
...
+
+    o_Y->dims[0] = 1;
+    o_Y->dims[1] = outputChannels;
+    o_Y->dims[2] = outputSizeY;
+    o_Y->dims[3] = outputSizeX;
+
+    /* MALLOC OUTPUT TENSORS HERE */
+
+    //mallocTensorData(o_Y);
+    o_Y->float_data = (float*)malloc(outputSizeX * outputSizeY * outputChannels * sizeof(float));
```
nopeslide, line 167: `mallocTensorData` is an actual helper function :D
either use this one or add a `TRACE_FATAL`

nopeslide (commented):
looks good, almost there :)

src/operators/ai.onnx/ConvTranspose/11/execute_operator__ai_onnx__convtranspose__11__T_tensor_float.c
nopeslide: then we could also loose the function prototype.

Coderitter-GmbH: good catch

Coderitter-GmbH:
That should be it

nopeslide (approved):

nopeslide:
Nice. thanks again for this PR

PR #92, opened 2021-03-09
restructured headers and pb files

nopeslide:
I dislike having the generated protobuf files in `/src/pb` so I moved them to `/protobuf` and removed `/src` from the include path.

no `/include` & `/protobuf` contain all headers we need to build and interface out library from the outside.

@alrevuelta this is much cleaner, isn't it?


tests/benchmarking.py
```diff
@@ -25,6 +25,7 @@ def benchmark_model(id, path, io, n_inputs, n_outputs, n_runs=1):
         elif result == 0:
             print("[Warning] Measured time is 0. Perhaps trying to measure a very small time."+
                   "There is a known issue with time library in Windows.")
+            model_times.append(result)
```
alrevuelta, line 28: I wasn't storing result as its 0 here. Any reason behind?

alrevuelta (commented):
@nopeslide Great idea. The `src/pb` was a leftover from the early days of the project :). Left a comment but LGTM.

tests/benchmarking.py
nopeslide: you divide by the length a few lines later, that threw a DivByZero exception

PR #93, opened 2021-03-09
enum defined before typedef to ensure compatibility with c++

steczol:
In order to ensure compatibility with C++ compiler the definition of `enum operator_status` should happen before `typedef enum operator_status operator_status`.

You can find an example in the link below:
https://riptutorial.com/c/example/6564/typedef-enum


nopeslide (approved):
weird looking diff, but lgtm

